PR #26, opened 2011-07-10
Fixed error on mapping for :Tabularize /|

franciscoj:
Hi,

I had problems with vim complaining about `:Tabularize /|` not existing. After a quick check in `:h` I saw that including the bar in a mappings need scaping. That's what fixed the issue for me on my dotfiles franciscoj/dot-files@cb501c3, so I hope it helps for you too.

PS: Thanks a lot for sharing this :D


spf13:
Thanks for the fix!

PR #38, opened 2011-09-07
Install VCSCommand as a submodule and enable it

jakzal:


spf13:
Using Fugitive instead for the same functionality (git only, but that's all I use). If you need VCSCommand please add it to the local bundles config file.

PR #44, opened 2011-11-24
Fix directory typo when trying to build Command-T

bpizzi:
1- On my config (zsh), bootstrap.sh can't cd into $HOME/.vim/bundle/Command-t, because the directory spell with a 'T' and not a 't'. This PR fix the bug.

2- Added "set hidden": useful when switching buffer, because without it, you need to save the file before jumping to another buffer.

Thanks !


spf13:
Manually merged via cherry-pick because wouldn't automatically merge.

PR #63, opened 2012-03-14
source support_function.vim or you will meet the "JavaClassNameFromFilen...

lihuazhang:
...ame Not Found" error when you try to complete the java class snippet.


lihuazhang:
Hi, this cl has one issue. The command " source $HOME/.vim/bundle/snipmate-snippets/snippets/support_functions.vim" will fail if one has never install snipmate-snippets before.
But I don't know where  it should be put...

PR #91, opened 2012-05-24
Changed csapprox plugin load order to make it work.

skyeagle:
I don't know why but `csapprox` didn't work without this changes with my `wombat` colorscheme in _terminal_ vim.


skyeagle:
Hmm.. After I have tried to change colorscheme from vim command line it broke it again. It looks like `easymotion` conflicts with `csapprox`

spf13:
That's unexpected since their functionality greatly differs. Probably should ping the authors about this. If I had to pick between the two I'd go with easy motion

spf13:
Closing pull request to keep things clean. It's clear there is likely an issue here, but not that this fix is the solution.

PR #96, opened 2012-06-11
Removed mkview calls because they are breaking navigation commands in rails.vim

leh:
The use of mkview seems to break the navigation commands in rails.vim. Most uses of navigation commands creating those errors: "E345: Can't find file “xxx” in path"

A bugreport in the upstream project indicates that the calls to mkview are cause of the errors (https://github.com/tpope/vim-rails/issues/25).
Removing those fixed the problem for me.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -130,7 +130,7 @@
 
     " Ruby
         if count(g:spf13_bundle_groups, 'ruby')
-            Bundle 'rails.vim'
+            Bundle 'tpope/vim-rails'
         endif
 
     " Misc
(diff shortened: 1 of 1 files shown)


spf13:
While removing those lines fixes the rails plugin, it does so at the cost of no longer keeping state. I'm wondering if there is a better way to retain both functions.

spf13:
Pull request #127 provides an elegant fix and is merged in. Please follow the directions in that pull request.

leh:
Thanks!

PR #100, opened 2012-07-06
Add guifont format for gtk2

gerrywastaken:
Gvim doesn't appear to support the ":h" delimiter in the guifont setting and instead requires an escaped space "\ ". Because of this, when you load the default spf13-vim on Ubuntu 11.10 (and probably 12.04 although can't check atm) the spacing between each character is much larger due to it being unable to understand the guifont format.

This patch fixes that issue.

You can see a example of the problem on this stack overflow page: http://stackoverflow.com/questions/3539437/gvim-ubuntu-letter-spacing-consolas-font

Thanks for making spf13-vim :)

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -538,7 +538,11 @@
     if has('gui_running')
         set guioptions-=T           " remove the toolbar
         set lines=40                " 40 lines of text instead of 24,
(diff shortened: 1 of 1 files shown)


spf13:
Thanks Gerry!

PR #106, opened 2012-08-13
Use ack-grep in the Ack plugin when suitable.

amirh:
As the ack utility executable file in Ubuntu is ack-grep
Check for this name also (and not only for ack) before deciding whether to include the Ack bundle.

See this discussion: 
https://groups.google.com/forum/#!topic/spf13-vim-discuss/krzn-VdotFs%5B1-25%5D

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -42,7 +42,10 @@
         Bundle 'gmarik/vundle'
         Bundle 'MarcWeber/vim-addon-mw-utils'
         Bundle 'tomtom/tlib_vim'
-        if executable('ack')
+        if executable('ack-grep')
+            let g:ackprg="ack-grep -H --nocolor --nogroup --column"
+            Bundle 'mileszs/ack.vim'
+        elseif executable('ack')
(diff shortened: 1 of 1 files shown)


petdance:
A suggestion: Bundle the ack standalone with the package.  It's only 80K and a single file.  Then no need to rely on the user's installation.

xoxo,
Andy the ack author

spf13:
Looking into this a bit, seems like a good idea. Thanks for the suggestion Andy.  

More on it here
http://betterthangrep.com/install/

Also seems like a better thing to bundle with the Ack vim plugin rather than spf13 itself.  

Best,
Steve

http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

petdance:
The Ack.vim plugin is something I don't maintain.  I don't disagree with your premise, though.

Honestly, I hardly use any vim plugins myself.  I live-and-breathe vim, but I think BufExplorer is about the only plugin I have.

Issue #73, opened 2012-04-10
Twig problem


stephpy:
Hi, i'm Symfony developer, then, i see we have a twig bundle on spf13-vim, but there is a problem,

When you type {{, it will do 
````{
}

```

Because AutoClose is mapped.

I did not see any possibility to override it to disable AutoClose for twig templates, have you an idea to solve this ?
```

**Comment by spf13** [comment]
we can just add  

```
  let g:autoclose_on = 0
```

to the twig file handler.

Best,
Steve

## 

Steve Francia
@spf13 (http://twitter.com/spf13)

jdewit:
A better solution would be to use the vim-twig plug http://www.github.com/beyondwords/vim-twig.git

Then just use the shortcut }} <tab> to render opening and closing twig brackets while still keeping the default autoclose functionality.

spf13:
I like this solution. Care to make a pull request?

Best,
Steve

http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

PR #115, opened 2012-08-28
added beyondwords/vim-twig bundle

jdewit:
Fixes issue #73

Check out the bundle at http://github.com/beyondwords/vim-twig.git


spf13:
These two commits are unrelated. 

Can you resubmit this pull request with only the first one and open a second pull request for the second commit?

PR #117, opened 2012-08-29
after bundle install, close all buffers.

glidenote:
Hi

when we do `:BundleInstall`, current version vundle create new buffer.
so we need close all buffers.


glidenote:
Thanks! :smile:

PR #118, opened 2012-08-29
Added beyondwords/vim-twig bundle

jdewit:
Still fairly new to contributing with github. Is creating branches to separate commits the way to go?

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -146,6 +146,11 @@
             Bundle 'tpope/vim-cucumber'
             Bundle 'Puppet-Syntax-Highlighting'
         endif
+
+    " Twig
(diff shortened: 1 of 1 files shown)


spf13:
Typically that's not necessary, just ideally each commit (and merge) would be isolated things. Could be multiple commits in a merge, but would expect them to all be related / depend on each other.

jdewit:
I understand why you'd want commits related to each other but new commits are automatically added to the existing pull request, you mean not pushing other commits until the pull request is merged? Either that or use a separate branch?

Thanks for putting this together btw. Been using vim for awhile and never knew about many of these plugins. Falling in love with vim all over again.

PR #119, opened 2012-08-29
Separate snippets to make it easier to override

jdewit:
I think the snippets bundle should be defined on its own so that it's easier to override with your own in .vimrc.bundles.local.


spf13:
I think this is a good idea, can you submit a new pull request within the updated file structure?

Issue #108, opened 2012-08-16
Tab to auto-complete with neocomplcache not working


jejansse:
Hey,

After a vanilla install of spf13-vim the tab-key does not seem to work with auto-complete. Even though it is set to behave the same as Ctrl-n, whenever I type tab to complete either nothing happens or a tab is inserted. The problem seems to be that snipmate is overriding the settings for tab, as can also be seen from `:map`, where tab is bound to a snipmate function. The strange thing is that even when I try to override the tab-completion in `.vimrc.local`, snipmate seems to be able to override it. When I override the tab behaviour in the vim command line after vim started up, I get the expected behaviour. I also am able to get the same behaviour when I remove snipmate.

Is there a way to make snipmate play nice with neocomplcache in spf13-vim? For example by letting snipmate use ctrl+k and disabling the snippets part of neocomplcache?

bronzehedwick:
I have the same issue with tab complete.

spf13:
I'm considering removing snipmate from the distro entirely and just going with neocomplcache. I'll do some more testing on this.  

Best,
Steve

http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

jejansse:
A temporary workaround while snipmate is still in the distro is to add the following in your `.vimrc.local`:

```
au BufEnter * inoremap <expr><TAB> pumvisible() ? "\<C-n>" : "<TAB>"
au BufEnter * inoremap <expr><S-TAB> pumvisible() ? "\<C-p>" : "<TAB>"
```

The second line is optional and only needed if you want shift-enter to go back in the completion list.

jdewit:
I like neocomplcaches snippet completion much better than snipmate. 

With snipmate removed theres no more conflict when completing snippets. 

+1 for removing snipmate

spf13:
This should be resolved now. You can no longer install snipmate alongside neocomplcache.

kc-dot-io:
When I hit tab to complete it just goes to the next item in the dropdown. If I hit enter on one of the options it doesn't complete and drops to the next line. Any idea what I'm doing wrong?

spf13:
It's in the readme under the neocomplcache section
https://github.com/spf13/spf13-vim#neocomplcache

<C-k> for completing snippets.

Does that answer your question?

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

huypn12:
[...]

PR #120, opened 2012-08-30
add neocomplcache-snippets-complete

jdewit:
Neocomplcache wasn't working for me and was telling me to install this bundle. 

Adding it solved the problem and I can now insert neocomplcache suggestions with 

``` html
<C-k>
```

I think this might resolve issue #108

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -88,13 +88,6 @@
         if count(g:spf13_bundle_groups, 'programming')
             " Pick one of the checksyntax, jslint, or syntastic
             Bundle 'scrooloose/syntastic'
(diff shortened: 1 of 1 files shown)


spf13:
I've resolved this using your approach, though letting people configure snipmate instead if they want.

phretor:
Looks like neocomplcache-snippets-complete is obsolete now.

```
% vim test
This is old version of neocomplcache-snippets-complete.
This is no longer maintained.
You should use neosnippet plugin instead of it.
https://github.com/Shougo/neosnippet
Press ENTER or type command to continu
```

PR #127, opened 2012-09-10
Allow users to disable views.

metcalfc:
There are several plugins that do not play well with views. Rather then
remove views entirely, allow users to disable them in their
vimrc.bundle.local. Provides an alternative to #96.


spf13:
Ok with this as a catch all, but probably better to automatically do this when breaking plugins are activated (with documentation).

PR #132, opened 2012-09-16
Fixed case of "leader"

krymen:
Fixed "leader" from upper to lower case in `.vimrc` file.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -350,7 +350,7 @@
     map <leader>et :tabe %%
 
     " Adjust viewports to the same size
-    map <Leader>= <C-w>=
+    map <leader>= <C-w>=
 
     " Easier horizontal scrolling
     map zl zL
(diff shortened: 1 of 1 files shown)


spf13:
Why do this? In almost all the documentation I've seen it's initial capped. Is there any benefit to doing this?

krymen:
Well, in my vim (on Ubuntu) initial capped doesn't work.

krymen:
I've just checked if `<Leader>` really doesn't work. It works only if I type `:source ~/.vimrc`. Same with `<leader>`. 

Seems that problem lies elsewhere. @spf13 Do you have any idea?

spf13:
Does it load the rest of your .vimrc file appropriately? Very odd that this one section would be ignored unless you source it.

Best,
Steve

http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

krymen:
Seems that everything else works after lunching vim.

Issue #133, opened 2012-09-17
Terminal not working after using the install script


mauserrifle:
I'm running Debian Wheezy and installing thru:

```
curl http://j.mp/spf13-vim3 -L -o - | sh 
```

at the end I get:

```
installing plugins using Vundle
Vim: Warning: Input is not from a terminal
```

Can't use the terminal after this. Seems stuck. CTRL-c does not work either. After closing my x-terminal I can work with vim tho, so the install seems successful.

Hitting some keys trying to exit gives:

```
maurice@maurice-pc:~$ maurice@maurice-pc:~$ maurice@maurice-pc:~$ maurice@maurice-pc:~$ bash: sdsadsa: command not found
                                    maurice@maurice-pc:~$ bash: sa: command not found
     maurice@maurice-pc:~$ bash: sada:q: command not found
                                                          maurice@maurice-pc:~$ bash: : command not found
                         maurice@maurice-pc:~$ > >
```

jhnwsk:
Hi,

I sumbitted a pull request fixing this issue for me.
If you like you can test it by creating creating an .sh script with the following code

``` vimscript
#!/usr/bin/env sh

endpath="$HOME/.spf13-vim-3"

warn() {
    echo "$1" >&2
}

die() {
    warn "$1"
    exit 1
}

echo "thanks for installing spf13-vim\n"

# Backup existing .vim stuff
echo "backing up current vim config\n"
today=`date +%Y%m%d`
for i in $HOME/.vim $HOME/.vimrc $HOME/.gvimrc; do [ -e $i ] && mv $i $i.$today; done


echo "cloning spf13-vim\n"
git clone --recursive -b vundle-terminal-input-fix http://github.com/jhnwsk/spf13-vim.git $endpath
mkdir -p $endpath/.vim/bundle
ln -s $endpath/.vimrc $HOME/.vimrc
ln -s $endpath/.vim $HOME/.vim

echo "Installing Vundle"
git clone http://github.com/gmarik/vundle.git $HOME/.vim/bundle/vundle

echo "installing plugins using Vundle"
vim -u $endpath/.vimrc_install - +BundleInstall! +BundleClean +qall
```

and running it thru some pipe of some sort like:
`cat the-script-file-you-created.sh | sh`

It worked for me.

cheers,
John

mauserrifle:
Hi John,

Works on my machine too :)

Closing this issue!

spf13:
Just yank it.

Steve Francia
[...]

Issue #93, opened 2012-06-03
problem installing in ubuntu


toutouastro:
When I copy paste this 

``` shell
curl http://j.mp/spf13-vim3 -L -o - | sh
```

I get this output and nothing happens :

``` shell
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   147  100   147    0     0    283      0 --:--:-- --:--:-- --:--:--   542
100   706  100   706    0     0    159      0  0:00:04  0:00:04 --:--:--   343
thanks for installing spf13-vim

backing up current vim config

cloning spf13-vim

fatal: destination path '/home/toutouastro/.spf13-vim-3' already exists and is not an empty directory.
ln: failed to create symbolic link `/home/toutouastro/.vimrc': File exists
Installing Vundle
Cloning into '/home/toutouastro/.vim/bundle/vundle'...
remote: Counting objects: 2349, done.
remote: Compressing objects: 100% (1511/1511), done.
remote: Total 2349 (delta 785), reused 2302 (delta 747)
Receiving objects: 100% (2349/2349), 283.96 KiB | 97 KiB/s, done.
Resolving deltas: 100% (785/785), done.
installing plugins using Vundle
Vim: Warning: Input is not from a terminal

```

kc-dot-io:
I am also seeing this issue.

toutouastro:
I tried running the command `curl http://j.mp/spf13-vim3 -L -o - | sh` from vim with `:!` but it didn't work too.

BafS:
I have the same issue

jhnwsk:
Hey guys,

I just posted a pull request #134 with an idea for fixing this.
#133 was a duplicate issue of this one I guess. You can check my [issue comment](https://github.com/spf13/spf13-vim/issues/133#issuecomment-8652911) there and see if it fixes the problem for you.

cheers,
John

spf13:
This should be resolved now. If not please reopen.

PR #134, opened 2012-09-18
vim terminal input fix

jhnwsk:
Hi Steve,

Thanks for the awesome vim distribution. It truly is the ultimate.

I saw some folks complaining about their terminals acting weird after running your install script with curl piped into sh so I gave it a little hack. Works perfectly for me now, hope you find it useful.

The error in bootstrap.sh causing the trouble is somewhat described here:
http://vim.1045645.n5.nabble.com/Vim-Warning-Input-is-not-from-a-terminal-td4332375.html

I also used parts of your .vimrc to start up vim in the script so that it does not prompt to "Hit Return" because of the `-` stdin mode. There is probably a ton of other stuff that can be removed from the .vimrc_install, I figured the Bundles info and 
`set shortmess+=filmnrxoOtT      " abbrev. of messages (avoids 'hit enter')` 
are useful but was not sure.

Thanks again.

cheers,
John


jhnwsk:
Oh, and sorry I haven't connected this to the issues I found - #133 and #93. I couldn't find info on how to do that. Hope you don't mind.

spf13:
I really like this idea, but worried it would be a maintenance mess with duplicate bundles. I'd rather move the bundles into another file and include them in both places. Perhaps I'll merge this in then do that.. Not quite ready to do that yet though.

jhnwsk:
Good point. This was kind of sloppy of me. I made some changes, don't know if you like the idea of having a separate file for the bundles but this would be enough for installer purposes. I think if extending the config is done with .vimrc.bundles.local then having a vimrc.bundles installed is kind of natural.

I tested this running a clean install and then running an update with .vimrc.bundles.local. Works perfectly although the installer is kind of ugly the first time around ;)

Not sure why github insisted on the last pointless merge-commit, but I still am a beginner.

PR #140, opened 2012-09-27
Allow overriding default snippets bundle without manually overriding the...

taxilian:
... whole section

This is useful for forks of spf13-vim so that we can provide a custom set of snippets for the subgroup using the fork without losing the ability to easily sync with the origin repository; even if you change the way you use the snippets we can still use our own.

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -102,7 +102,9 @@
     " Snippets & AutoComplete
         if count(g:spf13_bundle_groups, 'snipmate')
             Bundle 'garbas/vim-snipmate'
-            Bundle 'honza/snipmate-snippets'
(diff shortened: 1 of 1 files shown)


taxilian:
Either this solution or the [other pull request](https://github.com/spf13/spf13-vim/pull/119) work fine for me; whichever you decide to accept.  My primary reason for wanting to do it this way is to maintain backwards compatibility. With the other method anyone pulling that had done customization could lose their snippets if they don't update their .vimrc.bundles.local

spf13:
I think the other pull request is more ideal. I've asked him to resubmit within the new file structure.

Best,
Steve

## 

Steve Francia
@spf13 (http://twitter.com/spf13)

PR #143, opened 2012-09-27
Add support for .fork customizations

taxilian:
By adding a second tier of customization I can maintain some overriding options in a fork that I can use from multiple computers and share with my coworkers without needing to change anything in the core spf13-vim repo, making it much easier to remain in sync with the origin


spf13:
Can you add directions to the readme on how to use these?

PR #157, opened 2012-10-02
Add symlinks to .fork files in installer

taxilian:
Added the docs, but we didn't add links to the .fork customization from the installer =]

Unlike the .local files, these are intended to be in source control, just specific to a particular fork.


spf13:
please add to the windows install .cmd file as well :)

PR #161, opened 2012-10-06
Fix symlink check bug in dotfiles backup.

kew-l:
I think the following change to bootstrap.sh introduced in commit 2fd2644 to determine if the target is a symlink has a bug?

-for i in $HOME/.vim $HOME/.vimrc $HOME/.gvimrc; do [ -e $i ] && mv $i $i.$today; done
+for i in $HOME/.vim $HOME/.vimrc $HOME/.gvimrc; do [ -e $i ] && [ ! -L $file ] && mv $i $i.$today; done

On my macbook backups are not getting created for my pre-existing vim dotfiles.

I changed $file to $i and it works.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -22,7 +22,7 @@ echo "Thanks for installing spf13-vim\n"
 # Backup existing .vim stuff
 echo "backing up current vim config\n"
 today=`date +%Y%m%d`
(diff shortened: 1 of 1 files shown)


spf13:
Thanks Kevin.. Not quite sure how that typo made it in.

kevinSuttle:
Bump. This one still seems at large. I tried it twice and the install script didn't take into account that the `.vimrc` and other vim files in `$HOME` were already symlinked to my GitHub repo. The result was that nothing was getting backed up. Some output in the console for if/where the files are being backed up would be helpful.

PR #162, opened 2012-10-07
fix the conflict of <C-u>

JoveYu:
maybe <leader>u is more  appropriate


bronzehedwick:
Yeah, that's a better solution. Thanks!

Issue #167, opened 2012-10-11
After installing on windows, I get "Unknown function: fugitive#statusline" 


prongs:
http://www.evernote.com/shard/s134/sh/c95d28f9-46d1-44c6-ae71-3914f8bbfe9f/6071fa67a5a2fb2bbcc6e20fe8abc488

I am behind an http proxy. I changed git:// to http:// on line 5 in the easy install file(https://github.com/spf13/spf13-vim/blob/3.0/spf13-vim-windows-install.cmd). so that git clone can work just by setting environ variable http_proxy=proxy_address. I run the file via cmd, it succeeds and after that opening gvim gives this error.

jhnwsk:
Hi,

I had the same problem today. It seems that the .cmd installer script does not call 
`vim - +BundleInstall! +BundleClean +qall`. 

Just call this manually after installing and you'll have all the bundles. 
Worked for me on XP.

prongs:
I now get a different error:
http://www.evernote.com/shard/s134/sh/151ce676-cf67-4270-9861-4bc8a1a40468/3f31b5aaebcd46b3ab3ef06f5dd1b7e2

Any ideas?

On Thu, Oct 11, 2012 at 3:51 PM, John notifications@github.com wrote:

> Hi,
> 
> I had the same problem today. It seems that the .cmd installer script does
> not call
> vim - +BundleInstall! +BundleClean +qall.
> 
> Just call this manually after installing and you'll have all the bundles.
> Worked for me on XP.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/spf13/spf13-vim/issues/167#issuecomment-9335119.

## 

Rajat Khandelwal
Undergraduate Student
IIT Delhi

jhnwsk:
Seems like your bundles did not get installed at all. Did you see a vim window with bundles getting installed? If they fail they show an exclamation mark.

I am pretty certain Vundle uses https internally for downloading and installing bundles from github. If you can't clone using https this might fail.

You can also create a vimlog file for your installation process by running the installer with the -V9 option like:
`vim -V9vimlog.log - +BundleInstall! +BundleClean +qall`

This might give you some ideas.

jhnwsk:
Sorry, I might have mislead you. You get these errors when running the command not opening vim right? 

What you see means that you do not have Vundle properly installed. Did the part
`git clone http://github.com/gmarik/vundle.git %HOME%/.vim/bundle/vundle`
execute correctly?

[...]

PR #168, opened 2012-10-11
added BundleInstall to the windows installer .cmd file

jhnwsk:
Hi Steve,

I installed on XP today and found that for some reason the .cmd script does not call BundleInstall. After I launched vim it threw lots of red letters at me.

Fix is pretty simple. I tested on XP only.

cheers,
John


jhnwsk:
I think this behaviour is the reason for #167.

kew-l:
Wouldn't this need '-u "$BASE_DIR/.vimrc.bundles"' fed in as well?

call vim -u "$BASE_DIR/.vimrc.bundles" - +BundleInstall! +BundleClean +qall

spf13:
I'm traveling and won't be back to my windows machine for a couple days so really appreciate the help from others.   

This is accurate and should match the final call in the bootstrap.sh  

## 

Steve Francia
@spf13

jhnwsk:
Yeap you are right Kevin.

PR #171, opened 2012-10-12
windows xp installer .cmd

jhnwsk:
Since there is no mklink in windows XP and installing the tool that
provides this function is only possible on NTFS filesystems I created
this installer for my colleagues at work.

Probably far from a perfect solution but just so people on old XP boxes
can enjoy the awesomness easily.

What do you think?


jhnwsk:
I've added backup and update functionality to the xp-installer trying to mimic the behavior of bootstrap.sh for the most part. I replaced the part where you check if .vim stuff is -L before backing up with checking for an spf13-vim install. Not sure if it's the best idea but at least it reduces the wait for updaters. Tested on my xp box both fresh and update install and it works fine.

I think I do not fully understand the original windows installer. From what I see spf13-vim-windows-install.cmd does not have backup or update functionality. My "xp version" had only `mklink` replaced with `copy`. What am I missing?

PR #175, opened 2012-10-12
Add SQL, update README

jphustman:
I added some SQL Plugins that are handy.

To resolve conflict with Align `,tt` I changed the tagbar toggle to `<F9>`

I added numbers toggle to `<F3>`

Updated readme to reflect SQL and added some plugin documentation (and placeholders) with links.

I tried to condense the commits with `rebase -i` but it seems like it didn't work the way I expected it to.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -411,8 +411,12 @@
             \ 'file': '\.exe$\|\.so$\|\.dll$' }
      "}
 
+     " numbers {
+        nnoremap <F3> :NumbersToggle<CR>
+     " }
+
      " TagBar {
-        nnoremap <silent> <leader>tt :TagbarToggle<CR>
+        nnoremap <silent> <F9> :TagbarToggle<CR>
      "}
 
      " PythonMode {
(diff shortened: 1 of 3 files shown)


spf13:
First of all. I really appreciate the contribution. 

Unfortunately it's too much in a single pull request (and I don't want to merge all of it). 

Tabularize provides the same functionality as Align and in my opinion does it better. 

I'm happy to add the SQL plugins and welcome any updates to the readme... though if you look at the diff, it's very hard to follow what was changed. It seems that there were some unintentional deletions as well.  For instance, why not capitalize the first letter of a sentence? See 4837d385e1725108ad2bfa7aa66536403816d8a5 .

If it would be possible to break these into separate pull requests I'd be happy to merge nearly all of them in and provide feedback on anything I don't just immediately merge in. I could cherry pick, but I'd rather merge and give you all the credit. 

Did you look at tabularize at all? Do you think Align is better?

jphustman:
Ok Steve, I will close this request and make more smaller, simpler pull requests.

Align is required by the SQLUtilities plug in, but I will take a good look at tabularize and see if I can get it to do what I am looking for (or maybe somehow implement it's features into a forked SQLUtilities :)  I will test the two and see which I believe is a better plugin.

On the readme 4837d38, I capitalized a letter that got uncapitalized while editing.  The only way I can see changes to the README.markdown is if I commit and push it up, unless there is a better way to view it locally, I will google.

I tried to use GIT's rebase but I guess I need to learn it a little better.

PR #188, opened 2012-10-18
Disabled Annoying Relative Line Numbers By Default

konqui:
I think there is no really need for Relative Line Numbers, in case of development/debugging it is more annoying than useful.


spf13:
I'm sorry but this isn't disabling it by default, it's removing the ability entirely. I'd probably merge a commit that changed the default back to normal, but not one that removed this feature entirely.

PR #192, opened 2012-10-21
Switch from legacy NCC snippets to neosnippets.

jhenahan:
neocomplcache-snippets-complete is now a legacy repo. It has been replaced by neosnippet.


program247365:
FYI, for anyone seeing this, and getting an error in their VIM on startup, from the old vim plugin still sticking around, this worked for me:

``` bash
cd ~/.vim/bundle
rm neocomplcache-snippets-complete/ -rf
```

PR #207, opened 2012-11-10
Replacing Better-CSS-Syntax-for-Vim with vim-css3-syntax

jamestomasino:
Better-CSS-Syntax-for-Vim was conflicting with vim-haml on scss files. vim-css3-syntax seems to offer a similar level of css syntax highlighting without these issues.

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -152,7 +152,7 @@
     " HTML
         if count(g:spf13_bundle_groups, 'html')
             Bundle 'amirh/HTML-AutoCloseTag'
-            Bundle 'ChrisYip/Better-CSS-Syntax-for-Vim'
+            Bundle 'hail2u/vim-css3-syntax'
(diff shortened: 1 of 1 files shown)


ngaloppo:
Yeah, this seems to have made gVim infinitely slower, at least on my system. For example, loading a simple < 50 line Markdown file takes more than 7 seconds, loading html.vim three times!

```
9176.000  2688.000  2658.000: sourcing c:\users\ncgalopp\.vim\bundle\vim-css3-syntax\after\syntax\html.vim
...
11892.000  2656.000  2636.000: sourcing c:\users\ncgalopp\.vim\bundle\vim-css3-syntax\after\syntax\html.vim
...
14610.000  2658.000  2617.000: sourcing c:\users\ncgalopp\.vim\bundle\vim-css3-syntax\after\syntax\html.vim
```

PR #215, opened 2012-11-14
adding global variables to disable every bundle

jamestomasino:
This has come up before in several issues. I think it would be incredibly useful to have a variable capable of disabling any Bundle installed in spf13-vim. Since not every Bundle has an internal mechanism capable of doing this cleanly and they are all handled inconsistently when available, it seemed the best place to do this was in .vimrc.bundles.

I've kept a similar naming convention to existing disable variables.

```
g:override_spf13_bundle_{Bundle_Name_With_Underscore_Seperator}
```

For example:

```
g:override_spf13_bundle_vim_nerdtree_tabs
```

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -40,14 +40,24 @@
 
 " Bundles {
     " Deps
-        Bundle 'gmarik/vundle'
-        Bundle 'MarcWeber/vim-addon-mw-utils'
-        Bundle 'tomtom/tlib_vim'
+        if !exists("g:override_spf13_bundle_vundle")
+            Bundle 'gmarik/vundle'
+        endif
+        if !exists("g:override_spf13_bundle_vim_addon_mw_utils")
+            Bundle 'MarcWeber/vim-addon-mw-utils'
+        endif
+        if !exists("g:override_spf13_bundle_tlib_vim")
+            Bundle 'tomtom/tlib_vim'
+        endif
         if executable('ack-grep')
             let g:ackprg="ack-grep -H --nocolor --nogroup --column"
-            Bundle 'mileszs/ack.vim'
+            if !exists("g:override_spf13_bundle_ack")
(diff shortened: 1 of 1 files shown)


spf13:
I think this is a very interesting idea. I'd like to think about it a bit before merging it in. If we don't merge this in directly it will be something very similar based on it. Thanks for the commit.

spf13:
Thinking about this more, I like it, but the name seems incorrect. 'g:override_spf13_bundle_' ... override implies that we are providing something else instead. 

I think we should use the term 'g:spf13_disable_bundle_*' instead. I think it's much more descriptive. 

We also should add documentation for users to add this to their .vimrc.bundles.local file. 

I'm thinking there may be a way to automate this as well. I'm looking into it.

jamestomasino:
That's a great idea. I can update the variable names in my pull request branch if you'd like, or I can hold off if you think there's a way of automating the checks.

johntyree:
This has my vote. I've already had to replicate (and now maintain) two of these little bundle groups because I just wanted to remove one bundle from each.

What about a list of bundle names that should be disabled? Then you have one global variable and you can just do string comparison or something. The name strings are already well defined by `.vimrc.bundles` anyway..

spf13:
This is what I'm thinking of, just need time to do it.   

## 

Steve Francia
@spf13

spf13:
The commit 2d1b8f3ebb89b3fba8cda5b0bfa3075dddaa8fc1 has the same functionality but in a different interface. I feel that this interface is much better as it follows the same interface as Bundle already does and doesn't add any additional maintenance. 

It worked well for me, but please test it and let me know if it works for you. 

Important to note is don't add the UnBundle to your .vimrc.bundles.local as it loads too early and will just result in the bundle being re-enabled.

jamestomasino:
Absolutely brilliant. Thanks Steve

PR #218, opened 2012-11-15
MOD: map search key

e7h4n:
- make / search very magic like perl
- use g/ to search very no magic

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -87,7 +87,6 @@
             Bundle 'godlygeek/csapprox'
             Bundle 'jistr/vim-nerdtree-tabs'
             Bundle 'flazz/vim-colorschemes'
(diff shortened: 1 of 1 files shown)


jamestomasino:
Won't that cause a discrepancy for people using the search vs sed? If you get in the habit of / working in very magic mode, when switching to a sed command it becomes a bigger disconnect that now you aren't working in the same pattern mode.

I'd think this is more of a thing that belongs in a .vimrc.local. It could be very confusing remapping the default search. Have you considered making the first alias something more in line with your g/ that doesn't overwrite the default? Maybe something along the lines of v/ (though that would collide with visual mode)?

e7h4n:
@jamestomasino Vim very magic mode is more similar to perl regexp style like ack which I used. But, that's my personal habit, you are right.

PR #223, opened 2012-11-17
Added symbolic link creation for .vimrc.local and .vimrc.bundles.local

chipairon:
When installing/reinstalling using bootstrap.sh:

In case of people having a fork of this project and having added the customization files .vimrc.local and/or .vimrc.bundles.local to their repo, this changes will create a symbolic link similar to what is done to all other configuration files.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -36,12 +36,19 @@ fi
 
 echo "setting up symlinks"
 lnif $endpath/.vimrc $HOME/.vimrc
-lnif $endpath/.vimrc.local $HOME/.vimrc.local
 lnif $endpath/.vimrc.fork $HOME/.vimrc.fork
 lnif $endpath/.vimrc.bundles $HOME/.vimrc.bundles
 lnif $endpath/.vimrc.bundles.fork $HOME/.vimrc.bundles.fork
-lnif $endpath/.vimrc.bundles.local $HOME/.vimrc.bundles.local
 lnif $endpath/.vim $HOME/.vim
+
+if [ -e $endpath/.vimrc.local ]; then
+    lnif $endpath/.vimrc.local $HOME/.vimrc.local
+fi
+
+if [ -e $endpath/.vimrc.bundles.local ]; then
+    lnif $endpath/.vimrc.bundles.local $HOME/.vimrc.bundles.local
+fi
+
 if [ ! -d $endpath/.vim/bundle ]; then
     mkdir -p $endpath/.vim/bundle
 fi
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -40,6 +40,7 @@ lnif $endpath/.vimrc.local $HOME/.vimrc.local
 lnif $endpath/.vimrc.fork $HOME/.vimrc.fork
 lnif $endpath/.vimrc.bundles $HOME/.vimrc.bundles
 lnif $endpath/.vimrc.bundles.fork $HOME/.vimrc.bundles.fork
+lnif $endpath/.vimrc.bundles.local $HOME/.vimrc.bundles.local
 lnif $endpath/.vim $HOME/.vim
 if [ ! -d $endpath/.vim/bundle ]; then
     mkdir -p $endpath/.vim/bundle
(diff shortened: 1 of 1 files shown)


jamestomasino:
I thought that was the reason there was a .vimrc.fork and .vimrc.bundles.fork file. They local's weren't intended to be bundled into spf13-vim, but for people forking the project, the .fork files provided the exact same level of control without stepping on the toes of the .locals.

chipairon:
Take this response with a grain of salt, because I am new to this environment I have probably misunderstood how all the pieces fit together.

I would appreciate if you'd shed some light as to how could I achieve this:
    - Sync my ".local" configuration over multiple machines. 
Do you suggest making a fork and copying my ".local" files' content to ".fork" files? If there is a better way than to do a fork, please tell me.

I see your point but consider the case of people wanting to have its ".local" in a repo, so they can access them from everywhere they want. Why do you want to limit them to have ".fork" files in version control? Can't they have ".local" and ".fork"?

Imagine Foo as an awesome python and vim user and he makes a fork of spf-13 and customizes its ".fork" files greatly for his needs. Then Bar is in love with this fork and starts using it. Then Bar wants to have his own customizations in github, in his ".local" files so he can pull from Foo when he desires. What do you suggest Bar to do to be able to have his ".local" files synced in multiple machines?

That was one of the scenarios where I will find this useful. It doesn't hurt people not having their ".local" in the repo, but it helps people who do have them. Those ".fork" files where added to the picture 2 months ago (https://github.com/spf13/spf13-vim/commit/49525009d632239ac932d06581224a867fa0bd3d). I am pretty sure, there are some amount of forks prior to that date that use ".local" in their forks and would benefit. 

 Anyway, just 3 lines of code that don't add a lot, so no worries and thanks for the input.

spf13:
It's a good set of questions. I'll answer inline.   

## 

Steve Francia
@spf13

On Tuesday, November 20, 2012 at 4:48 PM, Chipairon wrote:

> Take this response with a grain of salt, because I am new to this environment I have probably misunderstood how all the pieces fit together.
> I would appreciate if you'd shed some light as to how could I achieve this:
> - Sync my ".local" configuration over multiple machines.  
>   Do you suggest making a fork and copying my ".local" files' content to ".fork" files? If there is a better way than to do a fork, please tell me.
>   This is exactly what fork is for.  
>   I see your point but consider the case of people wanting to have its ".local" in a repo, so they can access them from everywhere they want. Why do you want to limit them to have ".fork" files in version control? Can't they have ".local" and ".fork"?

What would be the point? Local is for changes local to that system. Fork is for changes in a downstream repo ( to be used on many machines )  

> Imagine Foo as an awesome python and vim user and he makes a fork of spf-13 and customizes its ".fork" files greatly for his needs. Then Bar is in love with this fork and starts using it. Then Bar wants to have his own customizations in github, in his ".local" files so he can pull from Foo when he desires. What do you suggest Bar to do to be able to have his ".local" files synced in multiple machines?

In theory we would need this. In practice we haven't. It's also against the idea behind this. The intent is to share and have a community developed distribution. I would certainly prefer in this situation to merge much of the good work Foo has done. If it is a dramatic
enough fork so that this wouldn't make sense then it wouldn't fit this model anyway and would take the place of master with bar being a fork of it.  

> That was one of the scenarios where I will find this useful. It doesn't hurt people not having their ".local" in the repo, but it helps people who do have them. Those ".fork" files where added to the picture 2 months ago (4952500 (https://github.com/spf13/spf13-vim/commit/49525009d632239ac932d06581224a867fa0bd3d)). I am pretty sure, there are some amount of forks prior to that date that use ".local" in their forks and would benefit.

Richard added them for the express purpose you outlined in your first question.

chipairon:
Thank you very much.

I understand it better, now. 

Discarding pull request in 3, 2, 1...

adamrights:
A few days ago I played around with turning .fork files into a submodule so I could checkout branches and swap tool sets quickly based on the project. I keep my two local files in a personal dotfiles-scripts-snips folder  and then symlink them to ~ where spf expects them. That way they're part of normal backing up.

PR #230, opened 2012-11-23
Don't read from stdin when bootstrapping.

johntyree:
I'm not sure why this was put in, but it's unnecessary and actually broke the
install process in my case. There's no input. Vim just hangs.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -50,4 +50,4 @@ if [ ! -e $HOME/.vim/bundle/vundle ]; then
 fi
 
 echo "update/install plugins using Vundle"
-vim -u $endpath/.vimrc.bundles - +BundleInstall! +BundleClean +qall
+vim -u $endpath/.vimrc.bundles +BundleInstall! +BundleClean +qall


jhnwsk:
Hi,

I put this in because of #133.
Without the fix, the installer causes the terminal to act chaotically after install is finished on my local machine. After installing on a remote machine via ssh: the session breaks alltogether.

In my case after starting vim said "no terminal input, reading from <stdin>" so I figured it was obvious to always read from stdin hence the `-`. Maybe I don't understand something @johntyree could you explain why your Vim hangs?

I am using Ubuntu 12.04 locally and some Debian remotely.

johntyree:
Because there's no input on stdin of course :)

I think the next question is where is the input you are seeing actually coming from?

If you look at the line that I changed in this patch, it should be fairly obvious that nothing is writing to stdin. Under normal circumstances, if you tell Vim to wait for input on stdin and then you don't give it any, it will just keep waiting until `EOF`.

@jhnwsk, What happens when you run `vim -u /dev/null -`  both locally and remotely? It _should_ say "Vim: reading from stdin..." and then hang forever.

I tried over ssh as well, both in a screen session and without it and they all do the same.

jhnwsk:
@johntyree `vim -u /dev/null -` hangs forever as expected. However from my understanding when you call vim from inside the bootstrap script it inherits the pipe as it's input source and vim really needs some kind of terminal. I am probably wrong but does this mean that my local and remote machines write something to stdin when calling vim and yours do not? So it's an environment issue really?

Anyway, the only alternative ways to do this I could come up with after intense googling are:
`vim </dev/tty -u $endpath/.vimrc.bundles +BundleInstall! +BundleClean +qall`
or

```
vim -u $endpath/.vimrc.bundles +BundleInstall! +BundleClean +qall
reset
```

or

```
gvim -u $endpath/.vimrc.bundles +BundleInstall! +BundleClean +qall
```

The last one sucks. What do you think?

jhnwsk:
Oh, and in naive self-defence I can also say that the original fix was inspired by [BestVimTips](http://vim.wikia.com/wiki/Best_Vim_Tips#Command_line_tricks) :)

johntyree:
If there is unread data on stdin then sure, vim will have something to read and be happy.

Again though, I'm not super interested in how to get vim to stop complaining. I'd rather know why you somehow have data coming in on `stdin`. Do we agree, at least, that there's no good reason for there to be any? Even if there were, vim is just reading it and then throwing it away.

It sounds like your environment is doing something tricky. Can you capture the input? Run `cat` instead of vim maybe? Or just change the script to call `vim -u /dev/null -` instead of that line and see if there's anything in the buffer?

jhnwsk:
Running `cat script | sh` where script is `vim -V9vimlog -u /dev/null -` produces the following vimlog:

```
XSMP opening connection
chdir(/dev)
fchdir() to previous dir
sourcing "/dev/null"
finished sourcing /dev/null
Searching for "plugin/**/*.vim" in 
...
sourcing some plugins which is not interesting for us
...
Opening the X display took 4 msec
Opening the X display took 2 msec
"-stdin-" 0 lines, 0 characters
XSMP handling save-yourself request
"-stdin-" 1 line, 1 character
"-stdin-" 1 line, 1 character
```

I'm not sure I understand this correctly but this `"-stdin-" 1 line, 1 character` come form the xsmp's SaveYourselfRequest upon freezing interaction with user. Vim itself shows nothing in the buffer.

I agree this behaviour is weird-ish. In my opinion using vim from a shell script like that is already a hack required to automate the vundle install process. If it were possible we would've used Ex mode, wouldn't we?

johntyree:
Why _can't_ we use ex mode? Seems to work, just isn't as beautiful.

But anyway, I don't think requiring that someone first download the file and then run it instead of (unsafely) running a script that are streaming down without having looked at it at all, is a bad idea.

Do you still have issues if you run the script like a sane person would? `sh bootstrap.sh`  ?

jhnwsk:
When called from within the installer as recommended in the documentation ex does nothing for me. I think it just doesn't go into display mode which is necessary for Vundle to run properly. Just guessing though.

When I run this with `sh` like a sane person probably would it works just fine.
On the other hand sane people tend to follow the documentation and `spf13`'s recommends curl for install and update.

@spf13 If you think messing up some terminals is cool then I am obviously also cool with that. The fixes with `</dev/tty` or `reset` are the only ones I can come up with for now. But since there already have been issues connected with this maybe a remark in the docs like "might make some terminals go bananas, if it does please reset" is the simplest fix. What do you think Steve?

johntyree:
I don't think there's anything really to be gained by using ex mode here. I do think, though, that if someone knows enough to be using vim and running scripts via `curl ... | sh` then it's not outrageous to say "If this fails because your terminal is wonky, then just download the script and run it normally."

`wget /tmp/... && sh /tmp/...`   is not difficult.

spf13:
We also could just force the installer to use /bin/sh instead of their
local shell.

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

johntyree:
To normalize the environment you mean?

`sh -c "curl ... | sh"` ?

spf13:
Yeah, something like that.

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

johntyree:
@jhnwsk can you try this instead?

```
sh <(curl http://j.mp/spf13-vim3 -L -s)
```

(I think this is bash only.)

jhnwsk:
@johntyree tried it, ran update fine. Terminal is a-ok afterwards. I also agree that a documentation only warning and `wget` alternative is probably enough. Just so there are no surprised faces out there.

johntyree:
@spf13, maybe change the 'howto' to use `wget` instead and just add a note that for bash-compatilble shells the way I suggested is a possible option? I think @jhnwsk is right about taking the policy of least surprise.

spf13:
I think this makes sense. Want to make a pull request?

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

PR #232, opened 2012-11-24
Ensure vim uses compatible shell in bootstrap

evidens:
If you're using fish or some other shell, this will keep the install from failing.
Should have no adverse affects for most systems.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -50,4 +50,7 @@ if [ ! -e $HOME/.vim/bundle/vundle ]; then
 fi
 
 echo "update/install plugins using Vundle"
+system_shell=$SHELL
+export SHELL="/bin/sh"
 vim -u $endpath/.vimrc.bundles +BundleInstall! +BundleClean +qall
+export SHELL=$system_shell


johntyree:
This is (possibly) about to become moot. See #230.

evidens:
As far as I understand it, regardless of which shell executes the install script, when running Vundle, the various installations will require shell interaction that assume a Bash-like shell.
Setting the SHELL variable to a known quantity avoids problems during the Bundle Install. 
This is a separate issue from #230 having to do with reading from STDIN.

johntyree:
Yes, but to solve the bizarre problem of a single character being pushed into `stdin`, it's possible that the shell will need to be forced to `sh` or some other lowest-common-denomiator during install.

Still trying to figure it out, was just giving you a heads up.

johntyree:
Also, I was under the impression that the $SHELL variable was supposed to tell about the capabilities of the shell, changing it is not going to somehow make fish into a bash-compatible shell if it wasn't before, no?

I think it might be more appropriate to actually just run the install in `sh` instead of the current shell.

Does 

```
sh -c "vim -u $endpath/.vimrc.bundles +BundleInstall! +BundleClean +qall"
```

work?

evidens:
The install script is _already being run under `sh`_. https://github.com/spf13/spf13-vim#linux-nix-mac-osx-installation

Changing `$SHELL` makes it so that shell commands invoked _from within Vim_ use the specified program. (It's not enough that Vim is being invoked from `sh`).

In the case where `.vimrc.fork` doesn't have `set shell=/bin/zsh`  (or other) this ensures all commands run in compatible mode.

Setting `SHELL=/bin/sh` is safe since the installation instructions already assume you are running the script under `sh`.

johntyree:
Ahh I see. I was mistaken about the semantics of the `$SHELL`.  Yes, I agree. This is a totally reasonable thing to do.

Issue #226, opened 2012-11-19
Snippet behavior unpredictable with code folding.


johntyree:
When I use python snippets (possibly others too?), I find that having `foldenable` set will complete destroy snippet editing process. Halfway into editing the snippet the code will automatically fold and then vim goes into visual mode for some reason, causing the code to be deleted.

I haven't thoroughly explored what's happening, but can anyone confirm?

```
vim test.py
i
classd<C-k>
(try to edit snippet)
```

My "solution" has been to turn code folding off, but there are still some strange conflicts with the Tab key, <C-k> and jumping between snippets. I'll dig deeper later, but for now I just want to know if it works normally for anyone/everyone else?

jphustman:
Confirmed, tab mapping left out with prior update.

johntyree:
@jphustman works great, but highlights a second problem.

When tabbing through the completion menu now, if you happen to land on a snippet it will end completion and expand immediately. 

Also, have we lost `<S-tab>` for going backwards through completion menu (like `<C-p>`) or did we never have it?

jphustman:
Thanks, I noticed some things that need to be tweaked as well on this very
useful plugin, and will work on getting a nice update over the weekend.
(Shuogo has been working hard :)

spf13:
Is this completely closed now?

johntyree:
I would say no.

Make a new file `vim test.py` and try the following.

```
clas<TAB>
```

Now the `class` snippet is highlighted, which is good. We can still `<TAB>` to get past it, which is also good. Instead, let's say we want to use it.

```
<C-k>  " snippet expands with name highlighted
SomeClass " Type the class name.
```

Now we want to jump to the second field. We press `<TAB>` but it inserts a tab character (which i think is the right choice.) So we try `<C-k>` instead.

```
SomeClass<C-k>
```

Now the code automatically folds up and becomes selected and we can't type anything at all. The incomplete class, with snippet markers, is hidden in the fold. If we unfold it manually, we can continue on with the snippet.

This doesn't seem to happen with C files, so maybe it's intereference from a python plugin?

At any rate, this is the default behavior right now and it's clearly not what anyone would want.

jphustman:
I think sometimes within a snippet, tabbing within the code may be necessary, but since we don't necessarily use tab's in spf13 (we use expandtab) I'm not sure, maybe more opinion is necessary.

[...]

PR #233, opened 2012-11-26
Tab S-Tab neocomplcache

jphustman:
scroll through list using tab s-tab, select and move through snippets using `<C-k>`
Issue #226


johntyree:
Looks like python-mode breaks this by rebinding `<S-TAB>` to do some strange autocompletion based on the Rope plugin.

PR #238, opened 2012-11-30
Make guifont work correctly in windows

jphustman:
This has never worked for me correctly in windows, but this does, please test.


johntyree:
Does this affect Mac users? Do they use gtk? I wouldn't know, just something to consider.

Isn't there a better way to determine the platform it's running on rather than checking font names?

spf13:
This commit breaks mac fonts, but I took your idea and made it work for linux, mac and windows in dda3fe463a83eda30fc0a6ed2a29ba413308fbb3 . Thanks for the contribution.

PR #247, opened 2012-12-06
Fixed echo printing out \n's

SecurityForUs:
Cosmetic but it is kind of annoying seeing "\n" in the output.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -17,19 +17,19 @@ lnif() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


spf13:
Kinda embarrassing this has been there so long. Thanks for the fix.

SecurityForUs:
No problem and it's all good.  Would've worked too by adding "-n" to it but why do that?  Lol.

Have a good one sir.

PR #250, opened 2012-12-10
Use more robust method for bootstrapping.

johntyree:
See #230.


johntyree:
No reason not to use `curl` directly.

spf13:
My comment didn't make a lot of sense.. conditional statements in a one liner. I ask forgiveness on account of travelling for the holidays and losing context. Still curl is more widely available so the last patch does it perfectly.

johntyree:
No harm done. Hope you had a nice holiday!

PR #256, opened 2012-12-22
Use invhlsearch instead of nohlsearch

johntyree:
This command toggles the highlight state, allowing the user to turn it off and
on as desired without it automatically turning back on.

I found it very annoying that some random command would set the search pattern
to something common like `\s\+` and suddenly my entire Vim session would be
highlighted. `<leader>/` turned if off but any kind of search related motion (or
running the command again, for example) would turn it back on again.

If the user turns off search highlighting, they probably don't want it to
immediately turn back on. If they do, it's as easy as `<leader>/` again.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -241,8 +241,8 @@
     nmap <leader>f8 :set foldlevel=8<CR>
     nmap <leader>f9 :set foldlevel=9<CR>
 
-    "clearing highlighted search
-    nmap <silent> <leader>/ :nohlsearch<CR>
+    " Toggle search highlighting
+    nmap <silent> <leader>/ :set invhlsearch<CR>
 
     " Shortcuts
     " Change Working Directory to that of the current file


jaywilliams:
I personally prefer the way spf13 currently handles this.

johntyree:
@jaywilliams, that's fair. I don't suppose everyone will agree on everything.

FWIW, my use case in particular was not actually related to my own searches, which you typically _do_ want highlighted. The problem comes when running commands such as those provide by latex-suite, which, during the course of execution, use the search feature for some obscure pattern that is really unrelated to what you want to do. Inserting environments, for example, results in all whitespace in the buffer being highlighted _every single time_ with no way to disable it.

I felt like this was a decent middle ground. I believe the only difference between this and the old behavior is that this doesn't automatically reenable highlighting if you've disabled it manually. Had you found something else or do you prefer automatic highlighting in all cases?

spf13:
I think this is fine. Makes it explicit without assuming the user meant something they didn't.

PR #277, opened 2013-01-26
Installation works in windows. Also, added number relative-absolute keymap 

prongs:
git clone doesn't work with git://. So changed it to https://. Also changed http:// to https://

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -529,6 +529,10 @@
         let g:indent_guides_enable_on_vim_startup = 1
     " }
 
+    "numbers {
+        nnoremap <F3> :NumbersToggle<CR>
+    "}
+
 " }
 
 " GUI Settings {
--- a/spf13-vim-windows-install.cmd
+++ b/spf13-vim-windows-install.cmd
@@ -2,12 +2,12 @@
 @if not exist "%HOME%" @set HOME=%USERPROFILE%
 
 @set BASE_DIR=%HOME%\.spf13-vim-3
-call git clone --recursive -b 3.0 git://github.com/spf13/spf13-vim.git "%BASE_DIR%"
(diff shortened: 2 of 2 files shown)


.vimrc
```diff
@@ -529,6 +529,10 @@
         let g:indent_guides_enable_on_vim_startup = 1
     " }
 
+    "numbers {
+        nnoremap <F3> :NumbersToggle<CR>
+    "}
+
```
jakzal, line 535: This change doesn't look like it should be part of this PR.

prongs, line 535: That is subjective but I think it should. You install the package `numbers` and don't provide any way to toggle b/w relative/absolute line number modes except for doing `rnu`  and `nu` manually. I had to go look for the function in the source that does the toggling. Also for people who were already using `numbers` in their vim configuration, if they decide to upgrade to `spf13-vim`, they have to re-enable the keymap.

jakzal, line 535: It is not related to the installation script and your PR is about that. I'm not negating the need for this mapping. I'm just saying it's not related to the current PR.

spf13:
This should be two pull requests. 

The first one (windows) one is perfect and should be merged in. It will help everyone.

The second one (mapping) I don't actually agree with. I'm pretty picky about what I allow for mappings to keys. I'd accept a mapping after the leader if it makes sense and doesn't conflict.

I also challenge how often people are turning on and off line numbers. In my experience people either nearly always use them or nearly never do. Optimizing this doesn't seem to be worth taking up a single key. How hard is :set nu! to type?

spf13:
I spent some time rewriting the Windows installer and included this change as part of that. Thanks for the feedback.

PR #292, opened 2013-02-09
Remove pymode's "set number"

rguillebert:
If pymode_options is set to 1 and "number" is unset, pymode overrides it and sets number for python files, that's why I set it to 0.


spf13:
Shouldn't this be 

let g:pymode_options_other = 0

rguillebert:
I don't think so :

https://github.com/klen/python-mode/blob/master/ftplugin/python/pymode.vim#L23

spf13:
I stand corrected. I was looking at an older commit.

Issue #255, opened 2012-12-22
<C-k> digraph binding overridden by neosnippet and easier tab/window movement


jamestomasino:
It appears you have remapped the <C-K> sequence that is used as the default digraph map by vim. The first place, around line 194, you are using a map for easy movement between windows. I'm not 100% on map vs map!, but I think this one might be safe because it should apply to command, visual, and select modes if I'm not mistaken.

However, around line 453 you are expressly overriding the digraph binding with a neosnippet_expand_or_jump command.

If this is absolutely necessary, would it be possible to rebind the digraph key to something else before overriding it?

iolloyd:
I am going to have to remap this also, but it would be great if it was remapped at a higher level in the actual plugin

jamestomasino:
I wonder if it's possible to organize the keymappings overall in a good standard way so they can be tweaked or overridden. spf13 has bindings all in place throughout the vimrc which makes sense from that file's perspective, but since customizations all get loaded afterward, they're already in place before our locals have any effect.

I don't want to suggest breaking things up into a bunch of smaller files if that can be avoided, but being able to hook into things at different parts of the loading flow might help. Or perhaps the keybindings being defined as variables that get assigned later... if you could patch in and change them. I'm not sure.

spf13:
It's a worthwhile discussion to have. I'm all for a way that works better
for everyone provided it doesn't add any complexity.

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

johntyree:
I'm also struggling with how to handle this. Some of the more aggressive maps I just had to unmap manually in `.vimrc.local`, but that's a pretty terrible solution.

iolloyd:
I think if we put some of the more aggressive maps in the .vimrc.local file by default, then it wouldn't seem such a bad thing. There is no avoiding writing the mappings and then overriding but it should be as transparent as possible.  

Best,
Lloyd

http://www.linkedin.com/in/lloydster
http://careers.stackoverflow.com/lloydmoore
http://github.com/lloydmoore

johntyree:
I think having the `.vimrc.local` file be entirely separate is a good thing, though. This is a tricky problem.

If we had a way to set some variables before .vimrc is called then it could be handled with conditionals, but that still leaves the (very awful) problem of other plugin writers thinking it's alright to claim whatever random mappings they want without any way to disable them.

jamestomasino:
[...]

PR #299, opened 2013-02-19
Add option to skip window navigation mappings and keep digraph key

jamestomasino:
Addresses issue #255 by allowing users to `let g:spf13_no_easyWindows = 1` and disable the `<C-J>` `<C-K>` `<C-L>` and `<C-H>` mappings, thereby not overriding `<C-K>` as the digraph key.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -191,10 +191,16 @@
     endif
 
     " Easier moving in tabs and windows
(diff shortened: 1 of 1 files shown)


spf13:
Great idea James. Thanks!

PR #312, opened 2013-03-06
Pretty font support for Windows

jcbwlkr:
- Syntax error in elseif

Using linux. Causes error:

Error detected while processing /home/jwalker/.vimrc:
line  556:
E488: Trailing characters:         else has(


spf13:
Oops.. Must have hit undo before committing. Thanks for the quick fix.

jcbwlkr:
No problem. I just happened to be updating my install at about the same time.

PR #313, opened 2013-03-06
vim-gitgutter highlighting

jcbwlkr:
Background color of sign column in gitgutter is distracting. Per
airblade's docs adding this command cleans it up

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -131,6 +131,9 @@
 
     set cursorline                  " Highlight current line
 
+    highlight clear SignColumn      " SignColumn should match background for
+                                    " things like vim-gitgutter
+
     if has('cmdline_info')
         set ruler                   " Show the ruler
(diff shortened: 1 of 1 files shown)


johntyree:
The csapprox plugin is setting the gutter color to light gray. I had this line in my `.vimrc.local` already, but was surprised to find it was already present in `.vimrc`.  This needs to happen _after_ csapprox has loaded for it to work. Did you not have this issue @jacobwalker0814 ?

jcbwlkr:
Yeah @johntyree this is probably not perfect as it is. In my own setup my .vimrc.fork sets color to ir_black and then calls highlight clear SignColumn again. I'm not sure what would be the best setup here for a vanilla spf13 install. I think it also might depend on whether you use gvim versus terminal vim and whether or not you have solarized colors installed.

johntyree:
We just need a way to clear it after cs-approx sets it. It could go at the end of `.vimrc` perhaps? I guess it just needs to be after the colorscheme is set?

PR #323, opened 2013-03-21
Disable the neosnippet preview candidate window. When enabled, there can...

mmikitka:
... be too much visual noise especially when splits are used.


spf13:
Thank you. This was driving me crazy last night as I was coding on a flight.

PR #333, opened 2013-04-05
Use vim-snippets instead of snipmate-snippets

benvds:
As discussed on:
https://github.com/garbas/vim-snipmate/issues/114#issuecomment-15894423

On another note. Spf what plugin is adding these weird characters in the gutter before my line numbers? Thx!

weird chars


spf13:
Some of those characters I've not seen before, but gitgutter is the plugin that's putting them there.

PR #290, opened 2013-02-08
Updating powerline bundle

lucascaton:

Issue #267, opened 2013-01-17
vim-powerline has been broken


pdehlke:
Lokaltog has decided to force people to stop using the vim-powerline plugin; it is now completely useless. See the glory of the latest revision:

https://github.com/Lokaltog/vim-powerline/blob/new-plugin/plugin/powerline.vim

Needless to say, this leaves new spf13 installs as of January 2 2013 broken.

_sigh_

invernizzi:
## Solution

Install the new powerline dependencies

```
pip install https://github.com/Lokaltog/powerline/tarball/develop
wget https://github.com/Lokaltog/powerline/raw/develop/font/PowerlineSymbols.otf
wget https://github.com/Lokaltog/powerline/raw/develop/font/10-powerline-symbols.conf
mv PowerlineSymbols.otf ~/.fonts/
fc-cache -vf ~/.fonts
mkdir -p ~/.config/fontconfig/conf.d/
mv  10-powerline-symbols.conf ~/.config/fontconfig/conf.d/
```

Configure spf13

```
cat >> ~/.vimrc.local << EOF 
UnBundle 'Lokaltog/vim-powerline'                                               
 python from powerline.ext.vim import source_plugin; source_plugin()  
EOF

cat >> ~/.vimrc.bundles.local << EOF
Bundle 'Lokaltog/powerline'
EOF
```

run vim, and execute

```
:BundleClean!
:BundleInstall
```

parrotbsd:
Appreciate the fix. That error message was driving me batty ;)

Lokaltog:
Sorry about the mess, I didn't realize that other projects depended on the develop branch of vim-powerline. It has been marked as deprecated for some time now and it's recommended that users of vim-powerline migrate to the new project. Note that the old project is still available, but you have to explicitly checkout the develop branch as this is no longer the default branch.

spf13:
I think I have a semi creative solution for this. I'm just going to create
a fork of the project from the dev branch and move the bundle to that.

I applaud his intent of moving to a single powerline to use across shell,
vim and other locations, it doesn't fit at all with our goals here of no
needless dependencies.

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

Lokaltog:
That's too bad, but I can understand it if you don't want to have Python as a dependency.

spf13:
I don't. However it's quite easy to use the python one if you decided to by
just unbundling this one and installing that separately.

## 

[...]

PR #336, opened 2013-04-09
update powerline dependence

Gohan:
redirect to right runtime path


Gohan:
same as @lucascaton 's #290 solution
solving problem #267

Gohan:
add gohan@2d58e6c PR.

PR #339, opened 2013-04-10
replace to new powerline

Gohan:
powerline replacement from #336 add fallback.

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -85,7 +85,12 @@
             Bundle 'kien/ctrlp.vim'
             Bundle 'vim-scripts/sessionman.vim'
             Bundle 'matchit.zip'
-            Bundle 'Lokaltog/vim-powerline'
+            if has("python") || has("python3")
+                Bundle 'Lokaltog/powerline', {'rtp':'/powerline/bindings/vim'}
+            else
+                Bundle 'Lokaltog/vim-powerline'
+            endif
+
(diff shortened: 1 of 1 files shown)


chrisnicola:
:+1: thanks this fixes an error coming from GitGutters too 

https://github.com/airblade/vim-gitgutter/issues/65

spf13:
Pretty certain powerline requires both python && python3, but I could be mistaken. Either way, this is the right approach.

ljani:
This is a bit problematic approach on Windows, because only gVim is built with python support. Vim (which is used for updating the bundles) says `:echo has("python")` `0` whereas gVim says `:echo has("python")` `1`. Also, `:version` differs between Vim and gVim. If I were to change `spf13-vim-windows-install.cmd` to use gVim, the system is unusable during the update, because the updater steals the focus all the time. But at least it works.

So I'm not sure, which is the right approach, and I'm looking for comments before making a pull request.

EDIT: The new powerline is broken on Windows anyway. It keeps spawning `git branch -l` windows all the time.
EDIT2: Whoever else faces this problem, you can use the old powerline properly with `let g:spf13_use_old_powerline=1` in `.vimrc.bundles.local` not `.vimrc.local`. Took me a while to find the proper place.
EDIT3: I just found out that the Windows version of Vim on vim.org is very old. As a result, I upgraded my Vim and gave the new powerline a try again - now it seems to be working nicely.

PR #345, opened 2013-04-15
Provide ability to revert back to the previous version of powerline

mmikitka:


johntyree:
I don't see why this shouldn't be an option, but why are you interested in preserving the old powerline plugin even when Vim has python support?

mmikitka:
@johntyree I encountered character rendering problems in my terminal, and I am also happy with the current implementation and not ready to switch to a brand new extension that does not have as much mileage.

johntyree:
You can set the characters manually to just use `|` like before. As for the missing features, _that_ is pretty annoying so far. Taking a peek at the powerline repo suggests that they are coming down the line eventually, though.

spf13:
This powerline shift has caused a ton of issues for people. I'm happy to let people override to use the more tried and true one.

johntyree:
It was annoying for me to get the fancy characters correct as well. Considering I don't use powerline in any other apps, I'd honestly rather the old one stay maintained. Hopefully it doesn't rot too quickly.

PR #363, opened 2013-05-02
Don't load myusuf3/numbers.vim unless vim >= 7.3

mimetnet:


spf13:
Can't merge a pull request with 311 commits. Please just send the last commit in a new pull request.

PR #388, opened 2013-06-25
Set sh as the default shell.

Gaelan:
Some people use non-POSIX shells. Vim has trouble with non-POSIX login shells. Patching `.vimrc` to fix this is no fun. `vimrc.local` doesn't work because it is loaded too late in the process.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -19,6 +19,7 @@
 
     " Basics {
         set nocompatible        " Must be first line
+        set shell="/bin/sh"
     " }
 
     " Windows Compatible {


Gaelan:
Thanks for merging this!

yapz:
I don't know enough about .vimrc to tell you more, but this merge caused spf13 to throw the following on my mac.  I've tried re-installing, clean installing to no avail.  Commenting out this line did resolve it.

Error detected while processing /Users/xxx/.spf13-vim-3/.vim/bundle/vim-preview/plugin/preview.vim:
line   94:
E91: 'shell' option is empty
E484: Can't open file /var/folders/v3/txbzlqk10v77jpcf0zprmxkr0000gn/T/veitkpK/
Press ENTER or type command to continue

comeonly:
I got same error. remove `"` no error. I didn't check on win or other *nix.

```
// set shell="/bin/sh"
set shell=/bin/sh
```

Gaelan:
I just posted a fix for this: #390. Sorry for breaking HEAD.

PR #391, opened 2013-06-30
Don't set shell to /bin/sh on Windows

ljani:
Don't set shell to `/bin/sh` on Windows. Setting the shell to `sh` on Windows breaks eg. Vundle. Shell should default to  `%comspec%` on Windows.. I'm not sure if there's any help setting it to `cmd.exe` explicitly, so there's no point touching it on Windows.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -19,7 +19,9 @@
 
     " Basics {
         set nocompatible        " Must be first line
-        set shell=/bin/sh
+        if !(has('win16') || has('win32') || has('win64'))
+            set shell=/bin/sh
+        endif
     " }
 
     " Windows Compatible {


teasp00n:
Agreed. My shell is actually /usr/bin/sh so I had to override this locally anyway. I think if people are using 'non standard' shells then they will probably be aware they need to make alterations in their local vimrc.

Gaelan:
@teasp00n The problem is, editing vimrc (not .local) makes updates harder.

teasp00n:
@GaelanS Sorry I should have been more clear, when I said local vimrc I meant vimrc.local.

Gaelan:
Ah. But vim was erroring (I assume) before vimdc.local got loaded.

teasp00n:
Oh OK. I didn't get that although I'm not using windows however /bin/sh is not where my shell executable is. In any case I don't think setting the shell in the .vimrc like that is really suitable across all environments. Ideally it would use the $PATH to find it and if people have custom setups, defer the configuration to the .vimrc.local.

spf13:
What if you put that line in your vimrc.bundles.local? That loads a lot

## earlier.

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

teasp00n:
@spf13 Adding `set shell=/usr/bin/sh` to my .vimrc.local did fix my issue by the way. Not sure if you were talking to me or not...

PR #396, opened 2013-07-05
Add sparkup for Zen Coding.

### LINKED ISSUE #392: Unable to use ZenCoding.vim
I'm at a loss to explain what's happening. I've installed ZenCoding.vim via Vundle, but I can't complete anything. I'm using the default <c-y> mapping for it but no action. I'm guessing at an incompatibility between plugins. Has anyone else got this to work?

Gaelan:
Sparkup's default mappings conflict with our NERDTree mappings. Will fix in later commit. Fixes #392.

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -196,6 +196,9 @@
             Bundle 'amirh/HTML-AutoCloseTag'
             Bundle 'hail2u/vim-css3-syntax'
             Bundle 'tpope/vim-haml'
+            if (has("python") || has("python3")) && !exists('g:spf13_use_old_powerline')
+                Bundle 'rstacruz/sparkup', {'rtp': 'vim/'}
+            endif
         endif
 
     " Ruby


spf13:
Happy to merge, but waiting on the later commit first... Is it coming?

Gaelan:
@spf13 I just don't know what mapping I should use.

johntyree:
@Gaelan This PR is dead in the water without a solution for the mapping conflict. Preferably one that doesn't surprise people that don't know or care about sparkup.

What does this actually do? What are the bindings for? Can they be ignored entirely?

Gaelan:
It's a binding to expand the zencoding syntax to full HTML.

.vimrc.bundles
```diff
@@ -196,6 +196,9 @@
             Bundle 'amirh/HTML-AutoCloseTag'
             Bundle 'hail2u/vim-css3-syntax'
             Bundle 'tpope/vim-haml'
+            if (has("python") || has("python3")) && !exists('g:spf13_use_old_powerline')
```
johntyree, line 199: Does this depend on powerline somehow? What about people using airline?

Gaelan, line 199: Sparkup uses python for multi-editor support, sort of like powerline does, so I copied the condition for powerline. I should probably get rid of the last condition.

johntyree, line 199: Please do.

PR #404, opened 2013-07-17
Fixed bundling of Puppet-Syntax-Highlighting.

teasp00n:
Bundle 'Puppet-Syntax-Highlighting' -> Bundle 'vim-scripts/Puppet-Syntax-Highlighting'


teasp00n:
I was having issues with puppet not finding a resource it needed that would cause vim to crash. Qualified the bundle with vim-scripts and now things appear to be stable. Was puppet-syntax-highlighting confirmed working before? I don't use it myself...

teasp00n:
It doesn't look like this is actually necessary. Gonna close this for now.

Issue #406, opened 2013-07-19
neocomplcache-snippets-complete needs to be updated


orsenthil:
During upgrade, I got this message.

This is old version of neocomplcache-snippets-complete.
This is no longer maintained.
You should use neosnippet plugin instead of it.
https://github.com/Shougo/neosnippet

PR #407, opened 2013-07-19
Fix for - Issue 406 neocomplcache-snippets-complete needs to be updated

orsenthil:
Issue raised here: https://github.com/spf13/spf13-vim/issues/406

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -83,7 +83,7 @@
     " In your .vimrc.bundles.local file"
     " list only the plugin groups you will use
     if !exists('g:spf13_bundle_groups')
-        let g:spf13_bundle_groups=['general', 'neocomplcache', 'programming', 'php', 'ruby', 'python', 'go', 'twig', 'javascript', 'haskell', 'html', 'misc', 'scala']
+        let g:spf13_bundle_groups=['general', 'neosnippet', 'programming', 'php', 'ruby', 'python', 'go', 'twig', 'javascript', 'haskell', 'html', 'misc', 'scala']
     endif
 
     " To override all the included bundles, put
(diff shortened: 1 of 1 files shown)


orsenthil:
I did a upgrade of spf13 , I got this message.

```
This is old version of neocomplcache-snippets-complete.
This is no longer maintained.
You should use neosnippet plugin instead of it.
https://github.com/Shougo/neosnippet
```

This patch is addressing that message. Does it help?

ljani:
I'm **not** getting the message, because there is no `neocomplcache-snippets-complete` bundle included. Only `neocomplcache` which is different, because `neocomplcache-snippets-complete` is split into `neocomplcache` and `neosnippet` afaik.

Then some off-topic talk:
On the other hand, at least I'm interested in replacing `neocomplcache` with [`neocomplete`](https://github.com/Shougo/neocomplete.vim), because according to the docs `neocomplete` is much faster than `neocomplcache` if Vim >= 7.3.885 and `+lua` are available. Many of the plugins already have the version requirement, but I'm not sure if any of them needs Lua. Thus I'm not sure if direct replacement of `neocomplcache` is welcome, so there probably should be some flags to choose either `neocomplcache` or `neocomplete`. It is a lot of work to upgrade the plugin and possibly do the flagging, and at least I don't currently have the time to do the transition. Hope someone will pick this up tho.

I guess I should've made a separate issue, but I don't want anyone to expect me to implement the transition anytime soon :)

spf13:
I don't know a single linux distribution that includes +lua by
default. Macvim doesn't either. This makes it something very
intentional. We could conditionally install it.

ljani:
Yes, I have exactly the same feeling. The problem is that `neocomplete` is not a drop-in replacement for `neocomplcache` and it needs all the same (but different) additional configuration. In case anyone's interested, there's this very incomplete [migration guide](https://github.com/Shougo/neocomplete.vim/wiki/neocomplete-migration-guide).

+1 for conditionality

spf13:
I'm closing this. It unfortunately looks like the error message you got referred to a previous state of the vim installation. Your patch doesn't quite line up with it either, you are actually removing the package it's telling you to switch to. It's an honest mistake, there's a lot of changes happening very quickly here and the names are all very similar.

orsenthil:
No problem. I did not quickly get back to this. But let me try it again and see to ensure that I do not stumble upon this issue.

PR #408, opened 2013-07-26
Replace vundle by NeoBundle

e7h4n:
NeoBundle features:
- lazy initialization for optimizing startup time.
- revision lock feature.
- other VCS(Subversion/Git).
- vimproc(asynchronous update/install).

And the latest update of vundle is 4 months ago.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -32,12 +32,14 @@
         endif
     " }
 
-    " Setup Bundle Support {
+    " Setup NeoBundle Support {
         " The next three lines ensure that the ~/.vim/bundle/ system works
-        filetype on
         filetype off
-        set rtp+=~/.vim/bundle/vundle
-        call vundle#rc()
+        if has('vim_starting')
+            set runtimepath+=~/.vim/bundle/neobundle.vim/
+        endif
+
+        call neobundle#rc(expand('~/.vim/bundle/'))
     " }
 
 " }
@@ -634,11 +636,11 @@
 " }
 
 " Functions {
-
     " UnBundle {
     function! UnBundle(arg, ...)
-      let bundle = vundle#config#init_bundle(a:arg, a:000)
-      call filter(g:bundles, 'v:val["name_spec"] != "' . a:arg . '"')
+        for bundle in filter(neobundle#config#get_neobundles(), 'v:val.orig_name ==# a:arg')
+            call neobundle#config#rm(bundle.path)
+        endfor
     endfunction
 
     com! -nargs=+         UnBundle
--- a/.vimrc
+++ b/.vimrc
@@ -40,6 +40,20 @@
         endif
 
         call neobundle#rc(expand('~/.vim/bundle/'))
+
+        command! -nargs=? -bang -bar
+              \ -complete=customlist,neobundle#complete_bundles
+              \ BundleInstall
+              \ call neobundle#installer#install('!' == '<bang>', <q-args>)
+
+        command! -nargs=+ Bundle
+              \ call neobundle#parser#bundle(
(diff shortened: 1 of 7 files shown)


spf13:
This is a great commit. It's also a huge change and will require a significant amount of testing. Thanks to an earlier pull request it also doesn't merge cleanly anymore, but that's hardly anything you could have done differently. I'm looking forward to taking some time to integrate this as it does appear that neobundle is an improvement.

e7h4n:
Conflicts has been resolved :smile:

spf13:
Thanks for resolving the conflict.. 

I've given this pull request a lot of thought. 

The question is what to do about the thousands of people who already have spf13-vim installed. I see two options. 
1. Add helpers so that Bundle and UnBundle work with NeoBundle.. Then everyones .vimrc.bundles.local files still work. People will continue to use Vundle until they run the upgrade script (just the installer again, but modified to actually upgrade) which will uninstall Vundle and add NeoBundle.
2. Make this a version 4.0 of spf13-vim. It's a pretty major change, considering Vundle is the core of spf13-vim today. I'm not against this approach, and these two options aren't exclusive. Perhaps we should do both.

Thoughts?

e7h4n:
A `Bundle` helper which compatible wich vundle is a good idea to make NeoBundle works well for vundle users, and `UnBundle` helper is already done.

The way to work of NeoBundle is very similar with vundle, so I don't think it's a good time to upgrade spf13-vim to 4.0.

Can we make a smooth upgrade?

X4:
I have reported a vundle bug, which until today wasn't answered or fixed, it's certainly very limited. NeoBundle looks really cool, but I guess @perfectworks is right, you either have to write a wrapper function, or you create spf14 branch and use the time to focus on spf14 and writing an upgrade script. I think the latter approach is less work, for both you and users.

e7h4n:
I've done some comparability work for NeoBundle, add fowllowing helpers `Bundle`, `BundleInstall` and `BundleClean`. I can use vundle command to manage my plugins by NeoBundle smoothly now, it works for me.

spf13:
I still need to do more testing on this. Thanks for the contribution. I just need to consider all upgrade cases before merging it in.

jphustman:
Just my two cents...  But I have a whole separate vim configuration (https://github.com/jphustman/cfml-vim) which is powered with NeoBundle and it works great.  There are definitely some differences that I believe would bump the version to a new release (4.0) with some added and refined documentation

I'd be happy to test and contribute to a 4.0 beta.

rinchen:
FWIW, I'm also +1 on NeoBundle and I would definitely use a beta based on it.

phcostabh:
+1

johntyree:
NeoBundle is obviously similar to Vundle, but this is a _huge_ fundamental change that might have all kinds of unforseen problems. I'm +1 on the change itself, but if this doesn't merit a major version bump, I think we can stop pretended "3.0" means anything at all.

ljani:
Btw, have you considered utilizing the lazy loading feature? For example load the Python related plugins only when editing Python files. Otherwise this looks very promising!

spf13:
https://github.com/spf13/spf13-vim/tree/nb

Here's a branch that we can use for this. I think it's probably the beginning of the 4.0 branch, but lets test a bunch first. We definitely need a solid upgrade strategy for all users (*nix & windows).

liming01:
I love the feature "lazy load". Now the performance of vim init slow down greatly after many plugins added.  Hope this feature added ASAP. Thanks.

FiloSpaTeam:
I switched to NeoBundle some times ago in my [fork](https://github.com/FiloSpaTeam/spf13-vim).
If you want, you can check ;-)

PR #410, opened 2013-07-29
added support for neocomplete

acgreek:
added support for using neocomplete

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -484,6 +484,108 @@
         nnoremap <silent> <leader>gg :GitGutterToggle<CR>
     "}
 
+    " neocomplete {
+        if count(g:spf13_bundle_groups, 'neocomplete')
+            let g:acp_enableAtStartup = 0
+            let g:neocomplete_enable_at_startup = 1
+            let g:neocomplete_enable_camel_case_completion = 1
+            let g:neocomplete_enable_smart_case = 1
+            let g:neocomplete_enable_underbar_completion = 1
+            let g:neocomplete_enable_auto_delimiter = 1
+            let g:neocomplete_max_list = 15
+            let g:neocomplete_force_overwrite_completefunc = 1
+
+            " SuperTab like snippets behavior.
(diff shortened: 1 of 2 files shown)


ljani:
Ah, just pointed this out in #407, many thanks!

ljani:
Meh, I did get my hopes up too early. So, it seems this is just copy pasted from the `neocomplcache` section and thus it's borked. If you take a look at [neocomplete.vim](https://github.com/Shougo/neocomplete.vim) setting examples section, you'll notice that for example auto start is set using `let g:neocomplete#enable_at_startup = 1`, not `let g:neocomplete_enable_at_startup = 1` as done in this commit.

So, to put it shortly, this does not work at all. Using `neocomplete` in place of `neocomplcache` needs more work than just search and replace the name.

@spf13, can you confirm that you read comments on closed pull requests?

EDIT: In case someone is interested in implementing a proper fix, some work has been put into the [migration guide](https://github.com/Shougo/neocomplete.vim/wiki/neocomplete-migration-guide) lately it seems. It's not full of empty cells anymore.

acgreek:
it does work for me, though it doesn't auto-popup by default. I just use CTRL-N when I want a completion. I could not use neocomplache on my code base because it's too big and vim was constantly blocking, searching for matches. What I did here was provide a functional base setup that allows one to use neocomplete. If you don't like the default setting, they can of course be changed as needed.

ljani:
I'm not saying it does not work :) Sure the script loads if it's listed in the bundles, but for example the auto-popup doesn't work, because the configuration is just wrong. I mean most of that configuration done in the commit is simply for nothing, because of the wrong variables that are set.

spf13:
Ill take a look later this evening, but in the meantime feel free to
fix and make a pull request.

ljani:
Ah, there was a comment already. Yeps, I followed the migration guide blindly and it seemed to be an easy job. Someone still should review the configuration.

The guide used to be all empty, but I noticed that it had been buffed lately, so it made the work much much easier.

PR #413, opened 2013-07-31
Airline: Configure to Look More Like Powerline

jcbwlkr:
I like the speed and lack of dependencies of vim-airline but I prefer the look of powerline. I added a theme "powerlineish" to vim-airline that mimics the colors of classic powerline. This commit enables that theme and enables the use of powerline fonts to get the nice separators.

I wasn't sure if this should go in `.vimrc` in the status line section or in `.vimrc.bundles`. I opted for the latter so the variables are only set if you're actually using that bundle. Additionally I considered adding a configuration variable to enable the powerline fonts but ultimately decided against it.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -158,6 +158,9 @@
         set statusline+=\ [%{&ff}/%Y]            " Filetype
         set statusline+=\ [%{getcwd()}]          " Current dir
         set statusline+=%=%-14.(%l,%c%V%)\ %p%%  " Right aligned file nav info
+
+        let g:airline_theme='powerlineish'       " airline users use the powerline theme
+        let g:airline_powerline_fonts=1          " and the powerline fonts
     endif
 
     set backspace=indent,eol,start  " Backspace for dummies
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -107,6 +107,8 @@
                 Bundle 'Lokaltog/vim-powerline'
             else
                 Bundle 'bling/vim-airline'
+                let g:airline_theme='powerlineish'
+                let g:airline_powerline_fonts=1
             endif
 
             Bundle 'Lokaltog/vim-easymotion'
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -107,8 +107,6 @@
                 Bundle 'Lokaltog/vim-powerline'
             else
                 Bundle 'bling/vim-airline'
-                let g:airline_theme='powerlineish'
-                let g:airline_powerline_fonts=1
             endif
 
             Bundle 'Lokaltog/vim-easymotion'


jcbwlkr:
@spf13 moved to vimrc

ngaloppo:
Is there a way we can check if the powerline patched fonts have been installed? Alternatively, can you suggest a .vimrc.local / .gvimrc.local approach that would let me use the same *.local files across systems that may or may not have have these fonts installed? Thanks!

spf13:
No way to tell from inside vim. Could do as part of the installer maybe,
but that doesn't handle the case where they install it afterwards.

As for the locals, you are talking about a Fork in spf13-vim terminology
and its totally supported.

Steve Francia
spf13.com
@spf13

On Aug 13, 2013, at 3:34 AM, ngaloppo notifications@github.com wrote:

Is there a way we can check if the powerline patched fonts have been
installed? Alternatively, can you suggest a .vimrc.local / .gvimrc.local
approach that would let me use the same *.local files across systems that
may or may not have have these fonts installed? Thanks!

—
Reply to this email directly or view it on
GitHubhttps://github.com/spf13/spf13-vim/pull/413#issuecomment-22547829
.

spf13:
The installer could potentially install the font as well, but it currently
doesn't do anything like that so it would be a lot of work to do that cross
platform.

Steve Francia
spf13.com
@spf13

On Aug 13, 2013, at 7:20 AM, Steve Francia steve.francia@gmail.com wrote:

No way to tell from inside vim. Could do as part of the installer maybe,
but that doesn't handle the case where they install it afterwards.

As for the locals, you are talking about a Fork in spf13-vim terminology
and its totally supported.

Steve Francia
spf13.com
@spf13

On Aug 13, 2013, at 3:34 AM, ngaloppo notifications@github.com wrote:

Is there a way we can check if the powerline patched fonts have been
installed? Alternatively, can you suggest a .vimrc.local / .gvimrc.local
approach that would let me use the same *.local files across systems that
may or may not have have these fonts installed? Thanks!

—
Reply to this email directly or view it on
GitHubhttps://github.com/spf13/spf13-vim/pull/413#issuecomment-22547829
.

ngaloppo:
In that case we should probably default to g:airline_powerline_fonts=0 in
.vimrc.

qva5on3:
If by default configuration would set 
`g:airline_powerline_fonts=0`
I suggest at least adding a comment explaining its purpose.

jcbwlkr:
@spf13 As others have mentioned above maybe we should turn the nice fonts back off. There have been a few issues filed recently from users without the fonts installed.

Perhaps we could disable the fonts but add a section to the README with screenshots showing what it looks like and how to enable them?

spf13:
I think that's a good idea. Pull request?

Steve Francia
spf13.com
@spf13

On Aug 21, 2013, at 8:28 AM, Jacob Walker notifications@github.com wrote:

@spf13 https://github.com/spf13 As others have mentioned above maybe we
should turn the nice fonts back off. There have been a few issues filed
recently from users without the fonts installed.

Perhaps we could disable the fonts but add a section to the README with
screenshots showing what it looks like and how to enable them?

—
Reply to this email directly or view it on
GitHubhttps://github.com/spf13/spf13-vim/pull/413#issuecomment-23013347
.

jcbwlkr:
I'll get on that.

jcbwlkr:
For those following this thread please reference the new PR #428.

PR #417, opened 2013-08-07
add multiple-cursors

bling:
this plugin is too awesome to be excluded from any distribution.  the only thing to worry about is that by default it maps `<c-n>`.

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -99,6 +99,7 @@
             Bundle 'tpope/vim-surround'
             Bundle 'spf13/vim-autoclose'
             Bundle 'kien/ctrlp.vim'
+            Bundle 'terryma/vim-multiple-cursors'
             Bundle 'vim-scripts/sessionman.vim'
             Bundle 'matchit.zip'
             if (has("python") || has("python3")) && exists('g:spf13_use_powerline') && !exists('g:spf13_use_old_powerline')


jcbwlkr:
I use this plugin in my personal setup. It's pretty sweet. My only problem with it is that in my experience sometimes feels kind of slow and doesn't catch every keystroke.

bling:
yep, terminal is much slower than [g/mac]vim.  but i've found that even taking into account typing a little slower it's often still faster than doing `s//` manually.

jcbwlkr:
@bling agreed.

e7h4n:
+1 for this, very awesome

spf13:
Thanks for the PR!
Can you suggest a logical remap that doesn't collide with core functionality and place it in the .vimrc?

bling:
hmmm, i thought `<c-n>` did something else by default, but `:h index` says it's the same thing as `j`.  i must of got it confused with insert mode's `<c-n>`.

spf13:
## Ah yes... So did I.

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

bling:
sorry, don't know where we left off with this.  do you want me edit the default mapping?  i figure that 99% of vim users probably don't use `<c-n>` when `j` is available, so i think it's a safe mapping change.

bronzehedwick:
Fugitive uses <c-n> to jump to the next commit to be staged in it's git
status window, which I actually use frequently. Would this remapping nix
that functionality?

On Fri, Aug 9, 2013 at 2:37 PM, Bailey Ling notifications@github.comwrote:

> sorry, don't know where we left off with this. do you want me edit the
> default mapping? i figure that 99% of vim users probably don't use <c-n>when
> j is available, so i think it's a safe mapping change.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/spf13/spf13-vim/pull/417#issuecomment-22414886
> .

## 

Chris
~
http://paradux-comics.com

bling:
@bronzehedwick it's safe, i use that all the time too.  fugitive sets up buffer local mappings which take precedence.

bronzehedwick:
@bling That's great, thanks for the confidence!

PR #428, opened 2013-08-24
Airline: Disabling Powerline Fonts and Updating README

jcbwlkr:
Makes the desired changes mentioned in comments of Issue #413.

Before merging there are a couple of items to address:
1. I added the Airline section at the end of the list of plugins. Perhaps it should be sooner so people are more likely to see it? I feel like the config loses some of its pizazz without the powerline symbols enabled.
2. I am hosting an example image on my own imgur account but it feels like it should be on the same account as the other screenshots. If you upload that image and give me the new link I will gladly change the README.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -164,7 +164,6 @@
         set statusline+=%=%-14.(%l,%c%V%)\ %p%%  " Right aligned file nav info
 
         let g:airline_theme='powerlineish'       " airline users use the powerline theme
-        let g:airline_powerline_fonts=1          " and the powerline fonts
     endif
 
     set backspace=indent,eol,start  " Backspace for dummies
--- a/README.markdown
+++ b/README.markdown
@@ -382,6 +382,14 @@ For example this screen shot demonstrates pressing `,,w`
 
 ![easymotion image][easymotion-img]
 
+## [Airline]
+
+Airline provides a lightweight themable statusline with no external dependencies. By default it uses the symbols `<` and `>` as separators for different statusline sections but can be configured to use the same symbols as [Powerline]. An example with and without powerline symbols is shown here:
+
+![airline image][airline-img]
+
+To enable powerline symbols first install one of the [Powerline Fonts] or patch your favorite font using the provided instructions. Configure your terminal, MacVim, or Gvim to use the desired font. Finally add `let g:airline_powerline_fonts=1` to your `.vimrc.local`.
+
 ## Additional Syntaxes
 
 spf13-vim ships with a few additional syntaxes:
@@ -466,6 +474,9 @@ Here's some tips if you've never used VIM before:
 [Matchit]:http://www.vim.org/scripts/script.php?script_id=39
 [Tabularize]:https://github.com/godlygeek/tabular
 [EasyMotion]:https://github.com/Lokaltog/vim-easymotion
+[Airline]:https://github.com/bling/vim-airline
+[Powerline]:https://github.com/lokaltog/powerline
+[Powerline Fonts]:https://github.com/Lokaltog/powerline-fonts
 
 [spf13-vim-img]:https://i.imgur.com/UKToY.png
 [spf13-vimrc-img]:https://i.imgur.com/kZWj1.png
@@ -475,3 +486,4 @@ Here's some tips if you've never used VIM before:
 [nerdtree-img]:https://i.imgur.com/9xIfu.png
 [phpmanual-img]:https://i.imgur.com/c0GGP.png
 [easymotion-img]:https://i.imgur.com/ZsrVL.png
+[airline-img]:https://i.imgur.com/sU0PG5P.png


barraponto:
but the `let g:airline_powerline_fonts=1` instruction isn't working for me (yes, i have the patched font and it used to work before this change).

mikedfunk:
Same here, I had to do this:

screen shot 2013-08-26 at 6 55 37 pm

devinus:
Airline seems to need `let g:airline_powerline_fonts=1` before being loaded to work. In Janus, I added this line to my `.vimrc.before`. I'm not sure what the analogous line is for spf13.

BenBergman:
@devinus try .vimrc.bundles.local

devinus:
@BenBergman Negative, doesn't seem to load before either.

spf13:
We probably need to add a .vimrc.before.

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

BenBergman:
Most of the stuff I have in my .vimrc.bundles.local would probably make more sense in a .vimrc.before anyway.

devinus:
Great, now I have no excuse to revert back to Janus. :(

spf13:
Wait, why is this a frown face? I figured this would help people in a lot
of ways, did I miss something?

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

devinus:
@spf13 It was a joke. :smile:

ljani:
Sidenote: at least following used to work from `.vimrc.local`.. Maybe there's some dark magic involved with loading the font.

```
unlet g:airline_powerline_fonts
let g:airline_left_sep=''
let g:airline_right_sep=''
```

jcbwlkr:
So the deal is that in Issue #429 we defined the left and right separators to be `›` instead of letting them default to `>`. The problem is that in airline if you specify a custom separator then that has precedence even over the `g:airline_powerline_fonts` setting. Removing the custom separators from `.vimrc` then following the README works to enable powerline symbols (although it doesn't show right away. It helps to type `:new` then `:bd` to force airline to update).

So how do we want to approach this? We can
1. Remove the custom separators. This makes enabling the fonts easier but looks worse for those who don't choose to enable it.
2. Add something like like `let g:spf13_airline_powerline_fonts = 1` to `.vimrc.before` that will toggle the appropriate powerline flags
3. Change the readme to say you need to add the following to `.vimrc.local`

``` viml
    let g:airline_powerline_fonts = 1
    unlet g:airline_left_sep
    unlet g:airline_right_sep
```

spf13:
I think the 3rd option. What makes spf13 good is that it just works.
Without any extra effort it should work. What makes it great is that it
retains the ability to customize things to your liking beyond the stock
install. Since power line fonts aren't installed with spf13-vim we
shouldn't expect them in the base, but make it easy to support them.

Steve Francia
spf13.com
@spf13

On Aug 28, 2013, at 8:32 AM, Jacob Walker notifications@github.com wrote:

So the deal is that in Issue
#429https://github.com/spf13/spf13-vim/issues/429we defined the left
and right separators to be
› instead of letting them default to >. The problem is that in airline if
you specify a custom separator then that has precedence even over the
g:airline_powerline_fonts setting. Removing the custom separators from
.vimrc then following the README works to enable powerline symbols
(although it doesn't show right away. It helps to type :new then :bd to
force airline to update).

So how do we want to approach this? We can
1. Remove the custom separators. This makes enabling the fonts easier
   but looks worse for those who don't choose to enable it.
2. Add something like like let g:spf13_airline_powerline_fonts = 1 to
   .vimrc.before that will toggle the appropriate powerline flags
3. Change the readme to say you need to add the following to .vimrc.local
   
   let g:airline_powerline_fonts = 1
   unlet g:airline_left_sep
   unlet g:airline_right_sep

—
Reply to this email directly or view it on
GitHubhttps://github.com/spf13/spf13-vim/pull/428#issuecomment-23410420
.

spf13:
After reading more carefully. 2 or 3 both do this. I have no strong
preference between them, but 2 seems easier ( 1 line vs 3 )

Steve Francia
spf13.com
@spf13

On Aug 28, 2013, at 8:32 AM, Jacob Walker notifications@github.com wrote:

So the deal is that in Issue
#429https://github.com/spf13/spf13-vim/issues/429we defined the left
and right separators to be
› instead of letting them default to >. The problem is that in airline if
you specify a custom separator then that has precedence even over the
g:airline_powerline_fonts setting. Removing the custom separators from
.vimrc then following the README works to enable powerline symbols
(although it doesn't show right away. It helps to type :new then :bd to
force airline to update).

So how do we want to approach this? We can
1. Remove the custom separators. This makes enabling the fonts easier
   but looks worse for those who don't choose to enable it.
2. Add something like like let g:spf13_airline_powerline_fonts = 1 to
   .vimrc.before that will toggle the appropriate powerline flags
3. Change the readme to say you need to add the following to .vimrc.local
   
   let g:airline_powerline_fonts = 1
   unlet g:airline_left_sep
   unlet g:airline_right_sep

—
Reply to this email directly or view it on
GitHubhttps://github.com/spf13/spf13-vim/pull/428#issuecomment-23410420
.

bronzehedwick:
My two cents would be for number 2, with added documentation about the option. 

```
—
```

mikedfunk:
Indeed, number 2 sounds like the preferred solution

## 

_Mike Funk_
mikefunk.com | mike@mikefunk.com

barraponto:
curiously, @jacobwalker0814 proposed solution doesn't work... in vimrc!
I mean, when I edit the vimrc file it uses `>`, but it uses the proper powerline separator for every other file...

jcbwlkr:
@spf13 I pushed an update to my branch that honors the option `g:spf13_enable_airline_powerline_symbols`

jcbwlkr:
@spf13 Did you get a chance to review this? Also do you want me to modify this to honor the changes made by @SaswatPadhi in #441? Finally would you prefer I open a new pull request?

spf13:
Yes to all questions. Thanks a lot.

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

PR #434, opened 2013-08-27
Source optional .vimrc.before

mikedfunk:
Adding an optional `.vimrc.before` file as suggested by @spf13 in Issue #428

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -15,6 +15,12 @@
 "   You can find me at http://spf13.com
 " }
 
+" Use vimrc before if available {
+    if filereadable(expand("~/.vimrc.before"))
+        source ~/.vimrc.before
+    endif
+" }
+
 " Environment {
 
     " Basics {


spf13:
This is great, and I'll merge it but could you follow up with an update to the docs so people know how and when to use it?

mikedfunk:
Sure, I'll get a new request on that today.

sbennett18:
Just curious. Do people think it would be good/beneficial to populate the `.vimrc.before` with all of the `g:spf13*` options commented out so you don't have to search around to see what can be easily changed?

mikedfunk:
@sbennett18 that would make it easy to know what is editable, but the `.virmc.before` file is not included in spf-13, it should be created by the user. So if you want to do that in your `.vimrc.before` file, go for it. (comment edited for clarity)

spf13:
Doing that will make it hard to update. We leave the local files missing so
that there aren't any conflicts when people make local changes.

Thinking through this more, this implementation is incomplete.

We need to load a .vimrc.before.local, a .vimrc.before.fork and a
.vimrc.before.

I'm fine with populating the .vimrc.before with all the options commented
with direct instructions to not alter this file, but copy it to
.vimrc.before.local and make modifications there. Then nothing gets out of
sync.

## 

Steve Francia
http://stevefrancia.com
http://spf13.com
http://twitter.com/spf13

spf13:
I've updated the code to reflect my comments above.

https://github.com/spf13/spf13-vim/commit/0240f11bfb17333ee01ed121036f000e66deeac4

We have no need for a .vimrc.before as we can just do it in the .vimrc file, so .vimrc.before.local & .vimrc.before.fork now exists which keeps it consistent with the other files that do similar things.

mikedfunk:
Makes perfect sense to me. I'll rename my before file accordingly.

PR #441, opened 2013-09-02
Conditionally override airline separators only if powerline fonts are not used

SaswatPadhi:
The Airline separators were unconditionally overridden. I have conditionally overridden the separators only if the user doesn't use Powerline fonts.

User is expected to specify `g:airline_powerline_fonts=1` in `~/.vimrc.before.local` instead of `~/.vimrc.local`.


sbennett18:
Thanks for doing this! I forgot to do something like this when I changed them initially.

PR #442, opened 2013-09-02
Added optionally bundling YouCompleteMe.

teasp00n:
So when I originally started writing this I was going to test for the binary like you suggested however I hit a bit of a catch 22. 

In order for the ycm_core to be available the plugin would have already needed to be bundled for the source to be available for compilation. As such it does not seem legit to check for the ycm_core before issuing the Bundle command or new users with fresh installs would never bundle it (without going out of their way to do it manually the first time to compile the core). 

I also didn't really like the idea of bundling it only to unbundle it straight after should the compiled core not be found. This approach could have left users confused as they might think they are using YouCompleteMe when it is in fact falling back to Neocomplcache or something similar.

At the moment my recommended approach would be to allow users to bundle it in their own .vimrc.bundles and then the onus is on them to then make sure the core is compiled. The plugin itself actually performs checks and tells the user if the core cannot be found. 

If you can think of a better way to do this I'm happy to make changes.

EDIT: I'm not really intending that you merge this straight away. Just wanted to start a discussion about it's implementation.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -520,6 +520,48 @@
         nnoremap <silent> <leader>ge :Gedit<CR>
         nnoremap <silent> <leader>gg :GitGutterToggle<CR>
     "}
+    
+    " YouCompleteMe {
+        if count(g:spf13_bundle_groups, 'youcompleteme')
+            let g:acp_enableAtStartup = 0
+            let g:ycm_collect_identifiers_from_tags_files = 1
+            let g:UltiSnipsExpandTrigger = '<C-k> <Plug>'
+            let g:UltiSnipsJumpForwardTrigger = '<C-k> <Plug>'
+            "let g:UltiSnipsJumpBackwardTrigger = '<C-l> <Plug>' " not sure
+            "what to set this to to be consistent.
+
+            " <CR>: close popup
+            " <s-CR>: close popup and save indent.
+            inoremap <expr><s-CR> pumvisible() "\<CR>" : "\<CR>"
+            inoremap <expr><CR> pumvisible()  : "\<CR>"
+
+            " Enable omni completion.
+            autocmd FileType css setlocal omnifunc=csscomplete#CompleteCSS
+            autocmd FileType html,markdown setlocal omnifunc=htmlcomplete#CompleteTags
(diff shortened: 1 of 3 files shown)


spf13:
I actually think this is a pretty solid approach. youcompleteme is not in the default bundle groups, so users need to opt in, but that's very easy to do. 

I think it will work pretty well. What's missing is clear directions how to enable YCM for users who want to. We should add it to the readme and website. 

Once that's done I'm happy to merge it all in.

teasp00n:
I'm happy to do this however I'm on holiday in Vietnam at the moment so won't be able to get to it until November. If someone else wants to do it in the meantime then cool, if not I'll do it when I get back. 

There are a couple of things I want to tidy up before it's merged anyway. The pop up menu will reappear after choosing a completion. Something to do with the pumvisible stuff I imagine. Are you able to have a look at that? If not then again, I'll fix it when I get home.

johntyree:
@teasp00n How is this looking ?

teasp00n:
I'll admit I'd forgotten about this. I can update the readme stuff but I'm not sure how to go about debugging the completion menu reappearing after a completion is selected. It looks like there is more than one completion engine registered on the same triggers.

johntyree:
@teasp00n that only happens in spf13? Do you have neocomplete/neocomplcache disabled when you test it?

teasp00n:
Both neocomplete and neocomplcache are not in use. The issue occurs in all filetypes that have completion enabled.

johntyree:
I'll try your branch and see if it behaves strangely here as well. Probably not tonight, though. Have you played with ycm outside of spf13? Does it still cause problems for you with a "clean" vim?

teasp00n:
Nah testing that is next on my list.

teasp00n:
OK so on a clean install of spf it appears to be OK. I'll sort my own life out at some point but the commit should be good. Just gotta update that README.

teasp00n:
Let me know if you want me to change anything Steve.

spf13:
Merged. 
I don't use YCM so we could really use others to test.

spf13:
Additionally NeoBundle supports building things upon install. We are working on switching to that in the nb branch. I would think the installer could be seamless using NeoBundle.

PR #453, opened 2013-09-07
Changed solarized options from high to normal.

teasp00n:


dtgay:
Big :+1: from me. I spent a good deal of time trying to figure out how to just disable the bright red/orange indent highlights, but when I found this fix I decided to keep them since they aren't eye-burning at normal solarization levels. :cocktail:

spf13:
It's a good commit, just doesn't cleanly merge, so it will take me a bit to apply it.

dtgay:
Perhaps you could just do it manually in the current `.vimrc`? Both lines are right next to each other still. You could still give @teasp00n credit in the commit message or pull request. :smile:

teasp00n:
Doesn't bother me. If it's going to be a hassle to merge it then feel free to re-implement it on HEAD.

PR #459, opened 2013-09-19
Add map for Tabularizing on left aligned commas

BenBergman:
Similar to the existing mapping to allow aligning (or Tabularizing) after left-aligned ':', this pull request allows the same for left-aligned ','.


jcbwlkr:
I've been doing this the hard way for so long... @BenBergman you're my hero!

PR #462, opened 2013-10-07
Add thinca/vim-fontzoom

ljani:
Added [`thinca/vim-fontzoom`](https://github.com/thinca/vim-fontzoom) as currently there seems to be no super easy way to control the font size. Typical use case is increasing the font size in case of a sudden presentation.

[`thinca/vim-fontzoom`](https://github.com/thinca/vim-fontzoom) includes a few default keymappings (`+`, `-`, `ctrl-scroll`), I added equals (`=`) mapped to reset. Vim 7.4 cannot map `ctrl-plus/minus/0` so the obvious mappings are ruled out. Let's hope Vim 8 changes this.

I also evaluated [`drmikehenry/vim-fontsize`](https://github.com/drmikehenry/vim-fontsize), but didn't like the separate font sizing mode. Others were more general font managers and those are overkills in my opinion. [`thinca/vim-fontzoom`](https://github.com/thinca/vim-fontzoom) seemed to be the best out there.

In an ideal situation, the window size of gVim wouldn't change, but I guess we have to live with this limitation.

Tested on Windows 8 x64 with vim 7.4. I guess someone should test this on some Linux distro before merging.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -791,6 +791,11 @@
         " https://github.com/airblade/vim-gitgutter/issues/106
         let g:gitgutter_realtime = 0
     " }
+    
+    " vim-fontzoom {
+        " vim-fontzoom already includes a few default mappings, add reset
+        silent! nmap <unique> <silent> = :<C-u>Fontzoom!<CR>
+    " }
 
 " }
 
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -126,6 +126,7 @@
             endif
             Bundle 'airblade/vim-gitgutter'
             Bundle 'tpope/vim-abolish.git'
+            Bundle 'thinca/vim-fontzoom'
         endif
 
     " General Programming


johntyree:
This is a pretty uncommon thing to use. I vote we don't give it any maps whatsoever.

Disable the default maps and put a commented out line in ~/.vimrc.before that would enable them. There are plenty of examples in there of what that would look like.

In general, I think we need less steamrolling and more user-configurable variables with sane defaults. That's what we're going for with all this, isn't it? It would be nice if updating spf13 didn't destroy my config everytime.

ljani:
Well, for me it's pretty natural to zoom things in and out and gVim doesn't provide any way to do this. Thus the need for the plugin.

In my opinion, there's really no point bundling a plugin without any default way to use it. Might as well leave it out and let the user add it themselves. I agree the mappings aren't the best, but Vim is the problem here. If any of you have better suggestions for the mappings, I'm happy to change them. But as said, without any mappings, the plugin is useless.

As usual, I opened this PR for as a request comments and so far you're the first one. Seeing it has been quiet here, I guess no one else needs to increase the size of the text from time to time.

johntyree:
Is there a problem with specifying your preferred maps in the ~/.vimrc.before file with a comment? It's _very_ easy to uncomment the line and activate them if someone chooses.

As far as I'm concerned, aggressively overwriting prime real-estate keys (without modifiers to boot) with mappings for a plugin that is only used once or twice per session is a bad idea. Do you find your self constantly zooming in and out?

It's a nice plugin and scratches your itch, which I understand, but you should play around with mappings that are less invasive and try to find one that balances things better. `<leader><leader>+` and `<leader><leader>-` for example?

ljani:
The problem with the maps in the `~/.vimrc.before` is that I don't consider it worth loading the plugin at all unless there's a way to use it. Opening gVim on Windows is slow already.

I agree that the mappings are poor, maybe the separate fontsizing mode would be better after all, but I'm not going to make a pull request for that, because it felt just awkward :) I actually do zooming quite often, but maybe it's just me and my work environment.

`<leader><leader>+` and `<leader><leader>-` are problematic, because you can't really keep tapping them repeatedly until you find a size that fits. Sure you can, but it's not quick enough any more.

Now that I've given this a little more thought, when someone at least said something (thanks for that), I realize the case is with gVim. I'm a gVim user, because I have to do work in a Windows environment, and gVim provides no functionality for zooming. Whereas most of you are probably on your terminals, which already provide such functionality. So, I guess majority of the users are better off without this. I'm going to install Cygwin and see if it eases things for me. On that note, I'll close this PR.

johntyree:
gvim is superior to terminal vim in every way except convenience and interacting with the shell.

I'd stick with gvim and put this in `.vimrc.bundles.local` if I were you.

ljani:
Yeah, that's what I did. Still evaluating Cygwin tho, zsh is so nice

johntyree:
You know... I don't know anyone who's bothered to switch to zsh and then
regretted it. I just can't bring myself to dick around with the config. I'm
sure it's worth it in the end but I dread it anyway.

2013/11/29 Jani notifications@github.com

PR #492, opened 2013-11-09
Remove useless help messages for TagBar and NerdTree

TylerBrock:
Both TagBar and NerdTree use precious screen space to display help messages, this turns n00b mode off.


TylerBrock:
bump

BenBergman:
I think these help messages should stay for the sake of new users. They are quicker references than the help pages. Perhaps what could be merged upstream is an option group where the end user can add a flag to their .vimrc.before.local and it will remove all unnecessary UI options. That way new users still get easy access to help while the advanced users don't have to configure several UI options themselves. This group could also remove the scrollbars and menubars in gvim.

TylerBrock:
cool

PR #493, opened 2013-11-11
Need to disable indent guides auto colors.

willdietrich:
If not explicitly disabled the defined colors are always overridden by auto.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -748,6 +748,7 @@
             let g:indent_guides_auto_colors = 1
         else
             " For some colorschemes, autocolor will not work (eg: 'desert', 'ir_black')
+            let g:indent_guides_auto_colors = 0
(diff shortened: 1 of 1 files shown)


johntyree:
Why are we using an autocmd here in the first place? How does this interact with users that want to hardcode their own indentguides?

willdietrich:
According to https://github.com/nathanaelkane/vim-indent-guides

That's the suggested way of overriding the default color settings. The problem is that unless we add this change the manually specified colors are never used because the plugin is active unless manually deactivated (the change I added).

johntyree:
I'm working on fixing some of the organic warts that have grown on ~/.vimrc
as people add things willy nilly. This is an example of one. Hopefully I
will finish it up tonight and it can start being tested. In the meantime,
this is probably ok.

2013/11/30 Will Dietrich notifications@github.com

PR #498, opened 2013-11-22
some useful configuration 

JoveYu:
add fullscreen mode for gvim, and change airline theme.

bufferline is a nice extension to show buffer in your statusline,which works well with airline.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -359,6 +359,9 @@
     map zl zL
     map zh zH
 
+    " fullscreen mode for GVIM and Terminal, need 'wmctrl' in you PATH
+    map <silent> <F11> :call system("wmctrl -ir " . v:windowid . " -b toggle,fullscreen")<CR>
+
 " }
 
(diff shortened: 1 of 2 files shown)


jcbwlkr:
Although I originally pushed for using the powerlineish theme in airline I have switched to the solarized theme in my own fork. As JoveYu mentioned in a commit message it matches the overall theme better.

There are a couple points to consider regarding making the airline theme switch:
- Screenshots in the readme will now be stale (staler?)
- Users have the option to use `vim-airline`, `vim-powerline`, or the newer `powerline` for the status bar and this change only updates the statusline theme if they use airline.

Just my two cents :)

ljani:
That fullscreen mode is not very portable, and as far as I know, spf13-vim aims to be cross-platform.

spf13:
While fullscreen isn't portable, it's also not the default. All this adds is a shortcut to do something you already can (or can't) do.

PR #503, opened 2013-11-29
Refactor

johntyree:
This PR organizes the various config files somewhat. The idea is essentially that `.fork` takes precedence over the standard files and `.local` should take precedence over all.

config-dependencies

This shouldn't change anything significant for the end user.


spf13:
Other than the observation above, I really like this and #505 and feel that both would really help things.

ljani:
IMO the installer/update script should honor overridden `g:spf13_bundle_groups`, because this feature mentioned in the readme:

> The simpliest (and safest) way to update is to simply rerun the installer. It will completely and non destructively upgrade to the latest version

The update script used to honor `g:spf13_bundle_groups` before https://github.com/spf13/spf13-vim/commit/08bce95a1d6b58d7ea78432ae7ae99385bd7f10f, but not `UnBundle`, because the command was not available in `.vimrc.bundles.local`. Having `UnBundle` in `.vimrc.bundles.local` is a welcome change.

I wonder which one would be preferable solution, if either:
1. Source something in the very beginning of `.vimrc.bundles`, before the conditional
2. Just remove the `-u "$HOME/.vimrc.bundles"` and let Vim source `.vimrc`

A manual `:BundleClean` afterwards does work no problems, but IMO the update script should take the groups into account.

johntyree:
I agree it's annoying to have the installer pull everything in only to throw half of it out. I think solution 2 is the right one. Just install the very basics and then run a standard update with `vim +BundleInstall +BundleClean +qall` and let it do what it was designed to do.

@ljani are you interested in fixing this?

ljani:
Sure, I can do a PR, shouln't be too hard. But first, I'm wondering why the installer is only loading `.vimrc.bundles` in the first place? Reasons I can think of:
- Speed, no need to parse the full RC file
- Avoid locking some files in the repositories. Does this happen? I remember having some problems with `:BundleClean` and [PIV](https://github.com/spf13/PIV), but I can't really recall what those were.
- Conflicting state of the repo and loaded plugins. Has anyone noticed problems with running `:BundleInstall!` or `:BundleClean` manually?
- Just a cool idea back then, but now considered legacy and causing problems :)

spf13:
Originally we did this because there were issues running the complete .vimrc file. It attempts to setup a bunch of things that aren't ready yet since the packages haven't been installed yet. There may have been other reasons, but I don't remember them.

ljani:
Okay, it's not a good idea to run `vim +BundleInstall +BundleClean +qall` when bootstrapping.

So.. `.vimrc.bundles.before`? There's at least configuring the bundle groups the powerline stuff that should be included there. Is this a good idea even with so few options? My home feels already a bit crowded with the `.vim*` files and folders.. Of course the best solution would be to detect and disable the "bunch of things that aren't ready yet since the packages haven't been installed yet", if the packages haven't been installed yet, but I'm afraid I'm not up for that job at least.

Thinking of alternatives.. A bootstrapper vim.rc, which is used from the install scripts, and would try to load _some.rc_ before `.vimrc.bundles`, but this is just littering the repo.

johntyree:
I agree that the number of files is at a maximum right now. I'd like fewer,
even, but what we have now works.

We need to source `.vimrc.bundles.before` for this to work. There isn't
really another good way. How that gets done is still up in the air. I made
myself a post-it to play with this somewhat and see if something reasonable
pops out. If not, then having a one-time penalty isn't _that_ bad, is it?

2014/1/7 Jani notifications@github.com

ljani:
Good to hear you're toying with it.

Yeps, it's not too bad to run the initial setup manually, but at least I'd prefer upgrading using the scripts to minimize my work. But yeah, I can live with manual updates too..

Brainstorming here.. What if the script created (extracted) an untracked and emptyish `.vimrc` and `.vimrc.bundles` into the home folder, which would call `.spf13-vim-3/.vimrc` and `.spf13-vim-3/.vimrc.bundles` in turn? I mean like turning the tables when loading the scripts. This way user could setup their own before and after stuff in those scripts.

johntyree:
What I don't like about that is that it encourages people to fork and customize and then work forever based on their dirty branch. Ideally, you'd be able to bring your dotfiles with you, dump them on a fresh spf13 install and get your environment back. I am constantly changing dev environments and for me that's a huge feature.

PR #505, opened 2013-12-01
No clobbers and cleanup

johntyree:
This is a large PR unfortunately, but it cleans up a lot of accumulated warts.

It assumes that #503 is going to be merged.

It mostly adds conditionals around things might overwrite user settings or steamroll other plugins' or users' mappings (did I mention that I _hate_ that?)

There are few parts where I corrected things that were implemented strangely, so if someone was depending on that broken behavior, like testing for 'x' to determine platform, then there might be a visible change. It's for the best though :)


.vimrc
```diff
@@ -15,17 +15,23 @@
...
-    if filereadable(expand("~/.vimrc.before"))
-        source ~/.vimrc.before
-    endif
-" }
-
 " Environment {
 
+    " Identify platform {
+        silent function! OSX()
+            return has('macunix')
```
johntyree, line 22: Need to test that this includes macvim.

johntyree:
Give me this weekend to play with this a little more and make sure nothing crazy happens. I haven't tested it yet on a clean user with no `.local`. I'm not sure if something is broken and then fixed because I'm overwriting it, etc.

spf13:
John,
Is this PR ready? I'm excited to get these good changes in.

johntyree:
Oh sorry. Yes, I've been using this with success. It will hopefully be OK
for others.

2013/12/26 Steve Francia notifications@github.com

Issue #464, opened 2013-10-12
Add mappings for other g* commands


memotype:
`j` and `k` are mapped to `gj` and `gk`, which is awesome, but you should also include these mappings for consistent behavior, otherwise it's confusing:

```
noremap $ g$
noremap <End> g<End>
noremap 0 g0
noremap <Home> g<Home>
noremap ^ g^
```

spf13:
Great idea, Pull request?

lukedomanski:
Hi,

This is annoying when ':set nowrap' is on.

While gj and gk seem to collapse to the same behavior as j and k for wrap and nowrap, the same can not be said for g$ , g0, and g^ versus $, 0, and ^, where it STILL causes the cursor to only move to the first/last column of the display in nowrap mode.

Since the default $, 0, and ^ are overwritten, you then have to find some other way of jumping to the start/end of the nowrapped line.

@memotype - can you please fix this so that either:
1) you can optionally turn it off by putting a option into .vimrc.before, as is done for other features (easiest way)
2) the mappings toggle on and off when wrap mode is changed (harder, but most transparent and thorough)

lukedomanski:
This was so crippling to the way I use VIM that I tried to fix it myself.

With my limited vimscript experience I could not figure out a way to toggle these mappings when wrap is set or reset. All I could come up with was to check if wrap was set each time the mapping triggered, and alter the normal mode command executed.

The following seems to work, if you are happy with it, please include it::

``` vim
    " Same for 0, home, end, etc
    function! WrapRelativeMotion(key)
        if &wrap
            execute "normal! g" . a:key
        else
            execute "normal!" a:key
        endif
    endfunction

    noremap $ :call WrapRelativeMotion("$")<CR>
    noremap <End> :call WrapRelativeMotion("$")<CR>
    noremap 0 :call WrapRelativeMotion("0")<CR>
    noremap <Home> :call WrapRelativeMotion("0")<CR>
    noremap ^ :call WrapRelativeMotion("^")<CR>
```

lukedomanski:
Close?

memotype:
Sorry, I haven't been doing any coding recently. Life, etc. If some one
else wants to take over these changes feel free.

gonzaloserrano:
I just run with this problem but something weird is happening.

Either i set wrap or nowrap, when i do a simple d$ the last character is not removed.

Anyone noticed this?

rekendahl:
[...]

PR #513, opened 2013-12-11
Changed g* wrap relative mappings to revert to standard behaviour under nowrap

lukedomanski:
As discussed in issue #464 the new g\* wrap relative line movement mappings still only move to the first or last column on the SCREEN when nowrap is set, when you would expect them to move relative to the LINE.

added/changed the following

``` vim
    function! WrapRelativeMotion(key)
        if &wrap
            execute "normal! g" . a:key
        else
            execute "normal!" a:key
        endif
    endfunction

    noremap $ :call WrapRelativeMotion("$")<CR>
    noremap <End> :call WrapRelativeMotion("$")<CR>
    noremap 0 :call WrapRelativeMotion("0")<CR>
    noremap <Home> :call WrapRelativeMotion("0")<CR>
    noremap ^ :call WrapRelativeMotion("^")<CR>
```

Please improve if you are unhappy with it, but as far as I can see, it does not break anything or slow down the movement noticeably.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -252,11 +252,19 @@
     noremap k gk
 
     " Same for 0, home, end, etc
-    noremap $ g$
-    noremap <End> g<End>
-    noremap 0 g0
-    noremap <Home> g<Home>
-    noremap ^ g^
+    function! WrapRelativeMotion(key)
+        if &wrap
+            execute "normal! g" . a:key
+        else
+            execute "normal!" a:key
+        endif
+    endfunction
+
+    noremap $ :call WrapRelativeMotion("$")<CR>
+    noremap <End> :call WrapRelativeMotion("$")<CR>
+    noremap 0 :call WrapRelativeMotion("0")<CR>
+    noremap <Home> :call WrapRelativeMotion("0")<CR>
+    noremap ^ :call WrapRelativeMotion("^")<CR>
 
     " The following two lines conflict with moving to top and
     " bottom of the screen
--- a/.vimrc
+++ b/.vimrc
@@ -252,19 +252,32 @@
     noremap k gk
 
     " Same for 0, home, end, etc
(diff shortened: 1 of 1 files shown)


lukedomanski:
Hold up...this commit does not handle the case that you give these g\* key strokes a motions in Visual mode. Either wait for a an update from me, or modify slightly if someone else beats me too it.

lukedomanski:
Code is now...I'm not game enough to say that is doesn't break anything again....so lets says it an improvement :)

``` vim
    function! WrapRelativeMotion(key, ...)
        let vis_sel=""
        if a:0
            let vis_sel="gv"
        endif
        if &wrap
            execute "normal!" vis_sel . "g" . a:key
        else
            execute "normal!" vis_sel . a:key
        endif
    endfunction

    " Map g* keys in Normal, Operator-pending, and Visual+select (over written
    " below) modes
    noremap $ :call WrapRelativeMotion("$")<CR>
    noremap <End> :call WrapRelativeMotion("$")<CR>
    noremap 0 :call WrapRelativeMotion("0")<CR>
    noremap <Home> :call WrapRelativeMotion("0")<CR>
    noremap ^ :call WrapRelativeMotion("^")<CR>
    " Over write the Visual+Select mode mappings to ensure correct mode is
    " passed to WrapRelativeMotion
    vnoremap $ :<C-U>call WrapRelativeMotion("$", 1)<CR>
    vnoremap <End> :<C-U>call WrapRelativeMotion("$", 1)<CR>
    vnoremap 0 :<C-U>call WrapRelativeMotion("0", 1)<CR>
    vnoremap <Home> :<C-U>call WrapRelativeMotion("0", 1)<CR>
    vnoremap ^ :<C-U>call WrapRelativeMotion("^", 1)<CR>
```

PR #516, opened 2013-12-19
neco-ghc moved from ujihisa to eagletmt

reem:
Small change to .vimrc.bundles to reflect that ujihisa no longer maintains neco-ghc and has transferred ownership to eagletmt. Just changed the source of neco-ghc from `ujihisa/neco-ghc` to `eagletmt/neco-ghc`


reem:
As seen here: https://github.com/eagletmt/neco-ghc/issues/23

spf13:
Thanks for this.

PR #518, opened 2013-12-27
disable neocomplete unless lua support

e7h4n:


spf13:
I commented on the PR itself.

JoveYu:
if vim not compiled with lua support, it will show a warning when you start vim.

PR #519, opened 2013-12-29
add vundle submodule point

s1341:
Make vundle a real submodule by initializing it to the current commit.


s1341:
Turns out this just introduces all manner of fail, because it means vundle can't update itself. Closing.

PR #529, opened 2014-01-14
Informations about how to add YouCompleteMe were wrong.

NoWiS-:


misterwayn:
Confirmed. Came here to report issue only to find this pull request.

PR #536, opened 2014-01-26
Added Chocolatey NuGet install instructions for Windows

switchspan:
Hi! I added the Windows instructions to the Readme.markdown file for using the [spf13.vim package](http://chocolatey.org/packages/spf13.vim) on the [Chocolatey NuGet](http://chocolatey.org/) site.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -54,13 +54,15 @@ If you have a bash-compatible shell you can run the script directly:
 ## Installing on Windows
 
 On Windows and \*nix [Git] and [Curl] are required. Also, if you haven't already, you'll need to install [Vim].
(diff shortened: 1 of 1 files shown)


spf13:
I followed these instructions on a brand new windows and chocolatey install. It installed vim 7.4 fine, but then when spf13-vim installer ran it complained it couldn't find 'vim'. 

I ran this from powershell.. I also tried with command, same result.

switchspan:
Steve -- thanks for the heads up. I will retry it with powershell on my end and see if I can determine what the issue is. I could pull my installation of vim out and use another chocolatey version of vim as a dependency, but I think it is based on Cream, which might create other issues.

rekendahl:
It would be really cool to have this work seamlessly. It's a great install option!

switchspan:
Steve, I updated the Chocolatey package and this should fix the vim dependency issues: https://chocolatey.org/packages/spf13-vim

PR #538, opened 2014-01-28
map `\` to `,`, becase <leader> takes `,`

e7h4n:


mikedfunk:
+1 I do this already in .vimrc.local

## 

_Mike Funk_
mikefunk.com | mike@mikefunk.com

johntyree:
Is this just for working around broken plugins that don't honor `<leader>` ?

e7h4n:
Accurately map `\` will not affect any maps like `\a` or `\b`.

MattWoelk:
Do you mean something like this? (From my .vimrc.local)

```
noremap \ ,
```

johntyree:
I'm totally against indiscriminately remapping a key like this.

Some plugins use `\` instead of `<leader>` which is annoying, but sometimes people (myself included) use `\` for other things. This kind of mapping belongs in your personal `.vimrc.local`.

What problem is this solving? Maybe I'm just not understanding you correctly.

spf13:
I agree with @johntyree Best in local.

PR #549, opened 2014-02-03
Fix clipboard error while running Vim in tmux

pitkley:
Running Vim whilst in byobu/tmux causes the clipboard not to work, failing with:

`E353: Nothing in register *`

This has already been discussed [here](http://stackoverflow.com/questions/11404800/fix-vim-tmux-yank-paste-on-unnamed-register), the [accepted answer](http://stackoverflow.com/a/11421329/758165) seems to solve it.

There probably should go more testing into this, maybe there are other edge-cases where this might break, but it worked for me both in tmux and directly in the Terminal (running Mac OS X).

Best Regards,
-Pit

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -71,9 +71,9 @@
     scriptencoding utf-8
 
     if has('clipboard')
-        if LINUX()   " On Linux use + register for copy-paste
+        if LINUX()         " On Linux use + register for copy-paste
             set clipboard=unnamedplus
-        else         " On mac and Windows, use * register for copy-paste
+        elseif $TMUX == '' " On mac and Windows, use * register for copy-paste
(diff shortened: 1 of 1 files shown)


pitkley:
I came to think about it, and there is at least one edge-case: tmux on Linux. I will try to test it, and report back, if this works or breaks something.

**Update**:
So it turns out that this is a tmux on OS X specific issue, and has nothing to do with Vim or this distribution directly.
The correct fix seems to be to use `reattach-to-user-namespace` and update the `tmux.conf`, as stated in several sources [[1](https://coderwall.com/p/j9wnfw), [2](https://github.com/ChrisJohnsen/tmux-MacOSX-pasteboard)].

PR #551, opened 2014-02-06
Add Shougo/neosnippet-snippets. Fixes #539.

### LINKED ISSUE #539: neosnippet-snippets cannot be found
Just install a fresh `spf13-vim`, get below error in vim.

```
neosnippet default snippets cannot be loaded.
You must install neosnippet-snippets or disable runtime snippets.
Press ENTER or type command to continue
```

Imgur

And it's odd that there's no such error in MacVim.

ljani:
[Shougo/neosnippet.vim](https://github.com/Shougo/neosnippet.vim) version 4.1 split the repository into two and the other half is now missing. This causes following errors on Mac (#539) and Cygwin at least:

```
neosnippet default snippets cannot be loaded.
You must install neosnippet-snippets or disable runtime snippets.
```

I'm not sure why gVim (on Windows) doesn't warn about this.

This PR adds [Shougo/neosnippet-snippets](https://github.com/Shougo/neosnippet-snippets) as we're already using [the "extra" snippets](https://github.com/honza/vim-snippets), so it makes sense to include the default snippets instead of disabling them. 

More info of the breaking change: https://github.com/Shougo/neosnippet.vim/commit/5dba6a034ed85fc6105f7aa052830cc39ad262bf

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -162,10 +162,12 @@
             Bundle 'SirVer/ultisnips'
         elseif count(g:spf13_bundle_groups, 'neocomplcache')
             Bundle 'Shougo/neocomplcache'
+            Bundle 'Shougo/neosnippet-snippets'
             Bundle 'Shougo/neosnippet'
             Bundle 'honza/vim-snippets'
         elseif count(g:spf13_bundle_groups, 'neocomplete')
             Bundle 'Shougo/neocomplete.vim.git'
+            Bundle 'Shougo/neosnippet-snippets'
             Bundle 'Shougo/neosnippet'
             Bundle 'honza/vim-snippets'
         endif
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -162,13 +162,13 @@
             Bundle 'SirVer/ultisnips'
         elseif count(g:spf13_bundle_groups, 'neocomplcache')
             Bundle 'Shougo/neocomplcache'
-            Bundle 'Shougo/neosnippet-snippets'
             Bundle 'Shougo/neosnippet'
+            Bundle 'Shougo/neosnippet-snippets'
             Bundle 'honza/vim-snippets'
         elseif count(g:spf13_bundle_groups, 'neocomplete')
             Bundle 'Shougo/neocomplete.vim.git'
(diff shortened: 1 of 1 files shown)


ljani:
Meh, managed to add them in the wrong order.. I guess it doesn't make much difference, but it's nice to have proper cosmetics.

johntyree:
In `.vimrc` we specify

```
            " Use honza's snippets.
            let g:neosnippet#snippets_directory='~/.vim/bundle/vim-snippets/snippets'
```

does this not handle this issue? I have no errors.

ljani:
[Honza's vim-snippets](https://github.com/honza/vim-snippets) is different from [neosnippet-snippets](https://github.com/Shougo/neosnippet-snippets). The line you quoted, enables Honza's snippets, see [Shougo/neosnippet](https://github.com/Shougo/neosnippet.vim) towards the end of the page 

@johntyree: What OS are you using? gvim on Windows doesn't whine either, but console vim in Cygwin throws the error.

johntyree:
@ljani I realize it's a different set of snippets, but I thought it would handle the "default snippets" problem. Perhaps not?

I'm on OSX and Linux, but my own spf13 is heavily modified and not a good reference ;)

ljani:
Ah, sorry, I misunderstood you. Anyway, the default snippets are loaded [here (+3 lines)](https://github.com/Shougo/neosnippet.vim/blob/5dba6a034ed85fc6105f7aa052830cc39ad262bf/autoload/neosnippet/init.vim#L46) whereas those additional snippets are only loaded [here](https://github.com/Shougo/neosnippet.vim/blob/5dba6a034ed85fc6105f7aa052830cc39ad262bf/autoload/neosnippet/init.vim#L66) after throwing the error.

The problem should be reproducible on OSX as well, as mentioned in #539. I guess your customizations solve the problem then.

lukedomanski:
@ljani "I'm not sure why gVim (on Windows) doesn't warn about this"'

Unlike you, I **was** getting this error on Windows gVim also. These commits solved the issue for me. Though I have not checked for side effects yet.

I don't understand what you an @johntyree were referring to above, do "we" need this commit? If so, why is it sitting so long without being merged?

Cross-ref: #539

PR #555, opened 2014-02-10
add app_dir var to point to root of application

### LINKED ISSUE #548: on changing the app base directory ($HOME/.spf13-vim-3)
I was about to submit a PR that added `base_dir` to bootstrap.sh set to `$HOME/.spf13-vim-3` by default, (to match the `%BASE_DIR%` env var in the windows install scripts) but it seemed like this had already been previously the case with the endpath variable and was removed in this changeset https://github.com/spf13/spf13-vim/commit/7677bdaf86af1be8f8af9db960de8ac2311a4f18 . Is there any reason we can't shouldn't have this variable as long as we keep this: `endpath="$( cd "$( dirname "$0" )" && pwd)"`?

ghost:
This should fix #548

We could still rename the other endpath variables, but i plan
on doing something with them in future commits.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 ############################  SETUP PARAMETERS
 app_name='spf13-vim'
+app_dir="$HOME/.spf13-vim-3"
 [ -z "$git_uri" ] && git_uri='https://github.com/spf13/spf13-vim.git'
 git_branch='3.0'
 debug_mode='0'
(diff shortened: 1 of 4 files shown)


johntyree:
Looks reasonable to me.

ghost:
well merge when you're ready then (unless you want endpath renamed now)

johntyree:
I'd like to test it at least once before merging, unfortunately that will have to wait for the weekend when I have some free time.

spf13:
I'm in the same boat as @johntyree .. might not even be able to get to it until next week. Sounds like one of us will get to it this weekend. Thanks for the solid contributions.

PR #556, opened 2014-02-10
Delete .gitmodules

ghost:
It doesn't seem like we need this submodule anymore since the script clones vundle itself.

is there a reason it's still here?


spf13:
I think this is probably just a legacy and harmless leftover. @johntyree Can you think of a reason?

johntyree:
I think you're right. We clearly get vundle manually in `bootstrap.sh`.

Issue #510, opened 2013-12-03
Lag when exiting insert mode


adambiggs:
When in insert mode, there is a 1s timeout lag after pressing `<Esc>` before returning to normal mode.

This caused by `<Esc>` key mappings found in these 2 places:
- [.vimrc](https://github.com/spf13/spf13-vim/blob/3.0/.vimrc#L332)
- [AutoClose plugin](https://github.com/spf13/vim-autoclose/blob/master/plugin/autoclose.vim#L92)

When these bindings are unmapped, exiting insert mode happens instantly without any lag.

biluncloud:
I have the similar problem, both when I press ESC from insert mode and visual mode to normal mode, the lag would appear. However when I run VIM using the vimrc without any plugins, the problem still exist. So I tried again and found when ESC is pressed in insert mode, the cursor remained unchanged for 1 or 2 seconds, however before the cursor could be changed, any further action was already permitted immediately after ESC is pressed. So did exit visual mode. (I am running gVim on Windows) .

johntyree:
I don't remember what I was doing at the time, but I have a comment in my `.vimrc.local` that says

```
" Map <ESC> to \033\015 in .Xresources. Now our Meta bindings don't
" conflict and <ESC> is instant! Beautiful!
```

That's linux only of course _and_ I'm not using autoclose so who knows.

ljani:
> I don't remember what I was doing at the time

[MinTTY's Tips](http://code.google.com/p/mintty/wiki/Tips) page has nice explanation:

> It's a historical flaw of Unix terminals that the keycode of the escape key, i.e. the escape character, also appears at the start of many other keycodes. This means that on seeing an escape character, an application cannot be sure whether to treat it as an escape key press or whether to expect more characters to complete a longer keycode.
> 
> Therefore they tend to employ a timeout to decide.

adambiggs:
I should probably mention that I already had settings in my `.zshrc` and `.tmux.conf` files to remove the lag at the terminal level, but Vim was still adding it's own lag because of these mappings.

ljani:
I finally spend some time with this.. Cygwin works fine with the [MinTTY's Tips](http://code.google.com/p/mintty/wiki/Tips) for remapping `ESC`, but that obviously doesn't work for gVim.

With gVim, changing `timeoutlen` to eg. 50ms helps, but I'm not quite sure if I want to do that. `ttimeoutlen` doesn't seem to have any effect with the current plugins.

[...]

PR #559, opened 2014-02-11
Screen program mapping is now opt-in (as it slows down exits from insert back to normal mode)

rekendahl:
To enable screen mapping add `let g:spf13_screen_mapping = 1` in your .vimrc.before

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -360,12 +360,16 @@
     " http://stackoverflow.com/a/8064607/127816
     vnoremap . :normal .<CR>
 
-    " Fix home and end keybindings for screen, particularly on mac
+    " Fix home and end key bindings for screen, particularly on mac
     " - for some reason this fixes the arrow keys too. huh.
-    map [F $
-    imap [F $
-    map [H g0
-    imap [H g0
+    " - __Note__ that this has been shown to slow down switching from insert
+    "   to normal mode (for users that don't use screen?)
+    if exists('g:spf13_screen_mapping')
+        map [F $
+        imap [F $
+        map [H g0
+        imap [H g0
+    end
 
     " For when you forget to sudo.. Really Write the file.
     cmap w!! w !sudo tee % >/dev/null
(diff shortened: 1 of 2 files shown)


.vimrc
```diff
@@ -570,7 +574,7 @@
         nnoremap <silent> <leader>gi :Git add -p %<CR>
         nnoremap <silent> <leader>gg :SignifyToggle<CR>
     "}
-
```
rekendahl, line 573: removed some annoying spaces as well to make the file not show syntax issues

rekendahl:
See #510 for discussion of this pull request

.vimrc.before
```diff
@@ -85,6 +85,13 @@
     "   let g:multi_cursor_quit_key='<Esc>'
     " Require a special keypress to enter multiple cursors mode
     "   let g:multi_cursor_start_key='+'
+
+    " Add mappings for Unix `screen` program.
```
johntyree: I know this comment is already long, but it doesn't say what this option actually does. What are we mapping here? Home and End? Does anyone know?

rekendahl: I have no idea. I was mostly interested in making them "go away" as they cause an almost second delay when exiting insert mode. 

I could easily see simply removing the whole thing! 

I don't think we should be focusing on external programs like screen and TMUX and instead leave that to the user.

adambiggs:
+1 for removing it entirely.

rekendahl:
Ok. Let me do a pull request removing it entirely. I vote for that too. We need to minimize what we have (spf-vim is on the upper end of content) and it seems like reducing assumption on external flows makes sense

rekendahl:
Ok. Updated pull request: https://github.com/rekendahl/vimeda/commit/9b2363616e46ee59ac4c0c0b513c989864471bef

Simply removes these mappings. If anyone is using them I hope the chime in

johntyree:
I followed the history back to where this was introduced. @taxilian if you're still using this, be aware that it's gone now.

PR #571, opened 2014-02-26
Add Twinside/vim-haskellFold

0x0L:
Better folding for Haskell:
https://github.com/Twinside/vim-haskellFold


kgadek:
+1. Did not see any issue while using it from `.vimrc.bundles.local`.

PR #578, opened 2014-03-09
Added bundle honza/vim-snippets for UltiSnips compatability

evanhorn:
UltiSnips is apparently now only a snippets engine without snippets. This simply adds the bundle honza/vim-snippets to the bundle list for youcompleteme such that honza's snippets are installed with UltiSnips.
This is the first time I've done a pull request, so if I'm not doing it right please let me know.


spf13:
Well done. Thanks!

PR #587, opened 2014-03-27
fix PyMode setting; add ctrlp-funky

JoveYu:
add ctrlp-funky; it fast and not depend on ctags

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -506,9 +506,15 @@
     " }
 
     " PyMode {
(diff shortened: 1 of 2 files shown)


mikedfunk:
The problem with ctrlp-funky is it only searches the current file for function definitions. Tagbar already takes care of this without ctags.

JoveYu:
bug it can help you jump to function more quickly

Issue #517, opened 2013-12-23
Pressing <s> causes some unexpected behaviour, instead of substituting


SamJoan:
Hi there!

I've downloaded the latest version of spf13, and I've run into something weird.

When I press s for substituting, instead of substituting the character directly below the character, a ">" prompt appears, as in the screenshot featured here

http://imgur.com/Qq92PgN

I am unsure what the functionality is, but it certainly doesn't replace anything. I think that it might be searching for two-character strings.

Please let me know if you are having problems reproducing, or if this is a feature and I am misunderstanding.

Thanks!
Pedro

BenBergman:
I just noticed this too. I think this is the result of the `vim-sneak` plugin. I find the hijacking of a standard vim command for a very unrelated action quite annoying. I think this should either be remapped to not clobber existing functionality or preferably be removed entirely. I haven't spent much time with `vim-sneak`, but I don't really see how it offers much of an improvement over `/ab<enter>`. I don't mind the extra keystroke as it allows me the flexibility to use a search other than two letters.

SamJoan:
I can confirm that adding 

```
UnBundle 'justinmk/vim-sneak'
```

and then `:BundleClean` solves the problem! In the meantime I found that another workaround is pressing v-c, which substitutes the current character, too.

@benbergman, thanks for the find. May I ask how you found out which plugin was using the map?

andru255:
thanks @BenBergman and  @droope, i have the same issue, but for that plugin what the advantage in the update for spf? should be a removed if is unnecesary for the improve the skills for the editor

spf13:
We should remove vim-sneak if it's causing issues. Or better remap that functionality to a non-conflicting letter. 

The advantage is that it can be used in motions with /ab<enter> can't. 

Often fA isn't enough and you really want fAB.  

Anyone care to make a PR?

JoveYu:
yeah before remap sneap conflicting letter, you should remove it!

PR #595, opened 2014-04-11
Removed bundle 'justinmk/vim-sneak'; it messed with default 's' key.

MattWoelk:
This relates to Issue #517 which was fixed in PR #570 but seems to need fixing again.


casimir:
:+1:

adambiggs:
FYI for anyone who misses this - `vim-easymotion` recently added a bunch of features heavily inspired by `vim-sneak` (activated by defining some extra mappings).

PR #605, opened 2014-05-03
Introduce .vimrc.bundles.default

ghost:
Hi, I think we should be able to overwrite "g:spf13_bundle_groups" for our forks in  ".vimrc.before.fork" file, so the user can install the fork's default bundles. Current bootstrap.sh uses .vimrc.bundles which does not source ".vimrc.before.fork". I have introduced .vimrc.bundles.default, which sources ".vimrc.before.fork" first, and then ".vimrc.bundles". In this way, we can configure the bundles as our fork needs during the installation.

Please check this out.

### DIFF
--- a/.vimrc.bundles.default
+++ b/.vimrc.bundles.default
@@ -1,0 +1,13 @@
+" Default Bundles {
+    " Use before config if available {
+        if filereadable(expand("~/.vimrc.before.fork"))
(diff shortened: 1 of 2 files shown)


spf13:
I like this idea. Can you send another PR with how to use this in the readme.

PR #624, opened 2014-06-17
check for plugin existence before setting options and mappings

nmacinnis:
Adds some checks to .vimrc for the existence of plugins before setting plugin-specific options and mappings. This makes it a little easier to figure out what key bindings are actually in use (when using :map and such) with a subset of the default bundles.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -427,13 +427,19 @@
     " }
 
     " PIV {
-        let g:DisableAutoPHPFolding = 0
-        let g:PIVAutoClose = 0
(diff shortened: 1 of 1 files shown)


ljani:
Works fine (ctrlp and nerdtree shortcuts at least) on Windows 8 x64 with Vim 7.4.169. I'm assuming you're worried about `if isdirectory(expand("~/.vim/..."))`.

spf13:
That I was.. Great, I'll merge it.

PR #629, opened 2014-06-24
use POSIX compliant escape sequence instead of \e

ghost:


ghost:
i'm working on a slightly more intensive refactoring, so i'm gonna close this one

PR #637, opened 2014-07-07
Don't conceal JSON syntax

### LINKED ISSUE #375: Why double quotes in JSON are hidden but show up only in current line?
As title. Double quotes in JSON are hidden and show up only in current line. What plugin it is? I would like to turn it off.

chrisnicola:
Because hiding characters is a terrible idea. I can't believe the JSON plugin actually does this by default but it does. This is apparently the proper way to disable it. Fixes #375.


ghost:
thanks for figuring this out!   saved me some time :)

PR #678, opened 2014-10-13
add jiangmiao/auto-pairs

taboularasa:
I was seeing errors suggesting that this was a missing dependency of something else. added it and errors went away, not sure if this is the correct fix but thought I would post it anyway.

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -156,6 +156,7 @@
             Bundle 'scrooloose/nerdcommenter'
             Bundle 'tpope/vim-commentary'
             Bundle 'godlygeek/tabular'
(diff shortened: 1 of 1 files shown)


bronzehedwick:
@taboularasa - Would you be able to paste in the errors you were seeing?

FiloSpaTeam:
I ve not seen your commit and i included same plugin in mine commit. 
I had the same problem with vimrc. the problem for me was a missing function that kill autoclose.
Replacing autoclose plugin with auto pairs solved the problem.
P.S.: auto pairs have more powerfull control in bigger inline double quotes and brackets ;-)

PR #691, opened 2014-11-20
Swapped vim-multiple-cursors

seyDoggy:
Swapped out terryma/vim-multiple-cursors for the better supported fork from kristijanhusak/vim-multiple-cursors

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -109,7 +109,7 @@
             Bundle 'spf13/vim-autoclose'
             Bundle 'kien/ctrlp.vim'
             Bundle 'tacahiroy/ctrlp-funky'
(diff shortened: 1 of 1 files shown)


ngaloppo:
Wouldn't we need to add the guards for NeoComplete as described here?
https://github.com/kristijanhusak/vim-multiple-cursors#info

bronzehedwick:
Good point @ngaloppo. Not sure if this will conflict with YouCompleteMe, but that might be worth looking into as well.

seyDoggy:
If we were to include those functions wouldn't it make sense to add full-on
support for neocomplete? As it is I still have to include it in my
`.vimrc.bundles.local`. I'd be game for full support.

PR #702, opened 2014-12-11
Update to ctrlpvim/ctrlp.vim

edi9999:
kien/ctrlp.vim hasn't been update for over 1 year, and the ctrlpvim version is working better, especially it will order the files intelligently, showing first exact matches


fabianhjr:
+1

PR #706, opened 2014-12-18
Introduce personal customizations

ootput:


fabianhjr:
I think the .local files are for each users' use for their personal preferences. :/

spf13:
The local files are meant for your hard drive alone. The fork files are meant for your fork. If you would like to make changes to the global then please put them in the .vimrc and .vimrc.bundle files.

ootput:
Bah, this was never the intention. Still navigating GitHub's functions. :)

Sorry for the trouble.
Joe

PR #712, opened 2015-01-02
Fixing bootstrap being destructive

MyChaOS87:
The upstream Version of  bootstrap.sh is destructive, as .vimrc.bundles.default only souces.vimrc.before.fork and not .vimrc.before or .vimrc.before.local. Now fixed sourcing .vimrc.before, which then sources the other two by default.

### DIFF
--- a/.vimrc.bundles.default
+++ b/.vimrc.bundles.default
@@ -1,7 +1,8 @@
 " Default Bundles {
+
     " Use before config if available {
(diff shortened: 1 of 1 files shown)


MyChaOS87:
Dear @spf13 i hope it is ok to mention you. I wanted to ask, what i should do to get this merged?

PR #714, opened 2015-01-07
Added some bundles

FiloSpaTeam:
Some usefull bundles that can helps with HTML, QML and others languages in general.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -442,9 +442,9 @@
             " Pencil {
                 augroup pencil
                   autocmd!
-                  autocmd FileType markdown,mkd call pencil#init()
-                                            \ | call litecorrect#init()
-                                            \ | call textobj#sentence#init()
+                  autocmd FileType markdown,mkd,md  call pencil#init()
+                                                \ | call litecorrect#init()
+                                                \ | call textobj#sentence#init()
                 augroup END
             " }
 
--- a/.vimrc
+++ b/.vimrc
@@ -81,6 +81,12 @@
     set mousehide               " Hide the mouse cursor while typing
     scriptencoding utf-8
 
+    " Autoreload the configuration every time
+    augroup reload_vimrc
+        autocmd!
+        autocmd BufWritePost $MYVIMRC source $MYVIMRC
+    augroup END
+
     if has('clipboard')
         if has('unnamedplus')  " When possible use + register for copy-paste
             set clipboard=unnamed,unnamedplus
--- a/.vimrc
+++ b/.vimrc
@@ -577,43 +577,65 @@
     " }
 
     " ctrlp {
-        if isdirectory(expand("~/.vim/bundle/ctrlp.vim/"))
-            let g:ctrlp_working_path_mode = 'ra'
-            nnoremap <silent> <D-t> :CtrlP<CR>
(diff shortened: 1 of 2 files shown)


bronzehedwick:
What's the use case for vim-project?

Also, while MatchTagAlways is very cool, it does have a python dependency. I think there is a match tag plugin without a python dependency that at least could be used as a fallback, if not the default.

bronzehedwick:
The match tag plugin without the python dep is gregsexton/MatchTag. Not as featured, but still does most of what MatchTagAlways does.

FiloSpaTeam:
I'll try gregsexton/MatchTag so i can have an idea how it works.

For vim-project: i can give you an example.
I've 3 directories for 3 different projects (for example 3 different companies) and you are working in all of those directories.
When you open vim (or gVim) you can select, with a number, in which tree you want work for and NerdTREE setup itself to this directory.

FiloSpaTeam:
Tryied and works very well. Changed default plugin for TagMatching.

bronzehedwick:
Cool. I can see the vim-project use case now. Thanks!

FiloSpaTeam:
You're welcome!

bronzehedwick:
What if we conditionally loaded the plugin, based on a check for python?

``` vimL
" Always highlight HTML/XML tags
if has('python')
  " With Python Deps
  Bundle 'Valloric/MatchTagAlways'
else
  " Without Python Deps
  Bundle 'gregsexton/MatchTag'
endif
```

FiloSpaTeam:
Good idea! I have forgotten conditional statement use XD

bronzehedwick:
:+1:

FiloSpaTeam:
Last commit make config more powerfull:
- Without bufferline now i can see easily the name of current buffer.
- To switch buffer i've insert BufExplorer, very good plugin.
- To have more easy like buffer change i have used tabline from airline (you can use <Tab> and <m-Tab>) :+1:

FiloSpaTeam:
Last commits include a lot of changes.
To have more control over frameworks like Laravel or Symfony we need php extended plugin that needs vimproc and Unite
So i:
- Replace ctrlp with Unite with <Leader>f for search in current directory and <Leader>r with recursive search with async option.
- Added more control over PHP with phpcomplete_extended for YouCompleteMe (anyone that use neocomplete can help me to config)
- Added 2 subgroups ('laravel' and 'symfony')
  **Why subgroups?**
  Because you need PHP to use like a charm.
- Created a config for Unite ctrlp.vim like
- Removed useless plugins like expand-region because this config work already

FiloSpaTeam:
Another set of changes. I worked with Markdown and rst, limelight and goyo to have better writing :).
Tryied to create a simple plugin for qt resources file (enables syntax highlight ).
Next commits try to give better refactoring support.
All of work it s possible because i think this is a very very beautiful configuration ;-)

FiloSpaTeam:
I think the configuration can be usefull.
I've done a lot of changes but works very well with and without gui.

ghost:
there are way too many changes in this one PR to be mergable. it must be split up into different pieces per added functionality

FiloSpaTeam:
ok so i create for each functionality a branch? :+1:

PR #721, opened 2015-02-04
Changed `<Leader>a=` to only align one `=` sign

jazmit:
Currently `<Leader>a=` acting on the following:

```
a = b == 3
xyz = u == v
```

produces

```
a   = b = = 3
xyz = u = = v
```

.. which is probably not what the user intended.
Since aligning `=` almost always refers to assignment, it makes sense to only align the first one.


bronzehedwick:
:+1:

PR #725, opened 2015-02-21
Add numbers.vim.

MatthewHager:
Numbers.vim should exist since it is referenced in the README and other places.


bronzehedwick:
I think this was removed due to popular demand, so I think the documentation referencing it should be removed instead. 

I personally use numbers.vim, but if I think a good amount of others were confused/didn't like it installed by default.

ghost:
agreed that the documentation should be removed instead.

spf13:
It's been removed from the readme. It's still in the website, but the site is overdue for an update anyway. I'll be updating the site in the next week or so and remove numbers then.

PR #741, opened 2015-03-29
Change symlinks to hardlinks in windows to avoid msysgit incompatibility

sumguysr:
Tested to work on Windows 8.1

### DIFF
--- a/spf13-vim-windows-install.cmd
+++ b/spf13-vim-windows-install.cmd
@@ -28,13 +28,13 @@ IF NOT EXIST "%APP_DIR%" (
 	call cd "%APP_DIR%" 
 )
 
-call mklink "%HOME%\.vimrc" "%APP_DIR%\.vimrc"
-call mklink "%HOME%\_vimrc" "%APP_DIR%\.vimrc"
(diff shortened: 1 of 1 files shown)


ghost:
so what's the incompatibility? I think we'd need verification from both windows 7 users and another windows 8 user before merging something like this. I have no windows boxen and i think @spf13 only has access to windows 7.

sumguysr:
Msysgit has chosen not to support symlinks, at all. 

http://osdir.com/ml/git/2013-08/msg00003.html

spf13:
@sumguysr what are the disadvantages of hardlinks?

gdw2:
Hardlinks worked for me on Windows 7.  And best of all -- they don't seem to require admin privileges!

ghost:
Hard links in POSIX systems can't cross filesystem boundaries. I don't know if this is the case for Windows.

ghost:
so what actually happens right now if you try to install it on windows? do you get a bunch of warnings? do the calls fail?

ghost:
It'd be cool to find a dedicated windows person to deal with any windows related issues.

ghost:
ping?

PR #742, opened 2015-04-01
Switch to neocomplete from obsolete neocomplcache

guersam:
From [neocomplcache repo](https://github.com/Shougo/neocomplcache.vim#neocomplcache)

> Note: It is not maintained well. You should use neocomplete instead.
> 
> https://github.com/Shougo/neocomplete.vim

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -91,7 +91,7 @@
     " In your .vimrc.before.local file
     " list only the plugin groups you will use
     if !exists('g:spf13_bundle_groups')
-        let g:spf13_bundle_groups=['general', 'writing', 'neocomplcache', 'programming', 'php', 'ruby', 'python', 'javascript', 'html', 'misc',]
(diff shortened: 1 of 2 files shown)


.vimrc.bundles
```diff
@@ -174,11 +174,6 @@
             Bundle 'Valloric/YouCompleteMe'
             Bundle 'SirVer/ultisnips'
             Bundle 'honza/vim-snippets'
-        elseif count(g:spf13_bundle_groups, 'neocomplcache')
```
guersam, line 177: Instead of just removing the whole section, we might keep it with a warning message for a while. It would be less surprising but someday we need to get rid of it anyway. What do you think?

ghost:
doesn't neocomplete require your vim to be compiled with +lua ? or is that something else?

EDIT: use the right name for the plugin!

spf13:
I agree with this, but we should leave the old one in for a bit longer I think.

PR #743, opened 2015-04-01
Adapt to last vundle changes

gonzaloserrano:


ghost:
I assume that spf13 breaks without this change due to bundle updates? I don't wanna break mine right now, so tell me a bit more about this, since @spf13 seems pretty busy.

gonzaloserrano:
Yep, https://github.com/gmarik/Vundle.vim/commit/356f245fbefa631aa54d2f8a65c05063b278dfe2 breaks spf13.

PR #745, opened 2015-04-02
Fix UnBundle command

ptzz:
Variable name changed in vundle, g:bundles -> g:vundle#bundles


ptzz:
Already fixed by  #743

PR #778, opened 2015-05-14
Add way to toggle background light or dark

cljoly:
There is maybe a better place for this.

I'm a VimL beginner, I hope my code is correct. At home, it works.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -75,6 +75,19 @@
 " General {
 
     set background=dark         " Assume a dark background
+
+    " Allow to trigger background
+    function! ToggleBG()
+        let s:tbg = &background
+        " Inversion
+        if s:tbg == "dark"
+            set background=light
+        else
+            set background=dark
+        endif
+    endfunction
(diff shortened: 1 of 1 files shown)


barraponto:
I like this.

cljoly:
Thanks!

barraponto:
@leowzukw @spf13 i like this, but I think `<Leader>b` is already used by python-mode.
maybe that's why it doesn't work in my install.

cljoly:
@barraponto `<leader>bg` would be a good option too, since it should not be used everytime.

barraponto:
@leowzukw yeah but when you type `<Leader>bg` you trigger the python mode `<Leader>b` before it gets the `g`. So, in other words, it doesn't work as is :-1:

spf13:
I don't do much python now a days so it worked for me. We could change it... but to what? Also what does <leader>b do in python? Should we change that?

cljoly:
@barraponto Why not `<leader>tbg`, if it is not used already.

Does it let anything free :smile: ?

PR #802, opened 2015-06-26
Remove .vim directory recursively

rogyar:
`~/.vim` is a directory, will not be removed without `-rf` parameters. Thanks!

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -13,6 +13,6 @@ die() {
 
 rm $HOME/.vimrc
 rm $HOME/.vimrc.bundles
-rm $HOME/.vim
+rm -rf $HOME/.vim
 
 rm -rf $app_dir


j1mr10rd4n:
If spf13-vim is installed using the current [installer script, line 116](https://github.com/spf13/spf13-vim/blob/3.0/bootstrap.sh#L116) creates a symlink from ~/.vim to ~/.spf13-vim-3/.vim; -rf flags shouldn't be required.

rogyar:
Have just checked the situation. The system already had `~/.vim` directory before running spf13-vim installation. In that case -rf flags might be useful. Thanks

spf13:
This worries me as you could do a lot of damage and the install script already moves the existing ~/.vim and replaces it with a symlink.

PR #808, opened 2015-07-25
Added Bundle 'powerline/fonts'

lewdlime:
Added Bundle 'powerline/fonts' since powerline, vim-powerline, & vim-airline won't really be usefull if they don't actually have fonts that make them look nice.


spf13:
thanks @laughingman182 !

PR #822, opened 2015-09-05
Fix .nvimrc symlink on bootstrap

agonzalezro:


bronzehedwick:
+1 yikes I made a bad merge

ghost:
I thought that was on purpose, since it should read nvimrc from there shouldn't it?

ghost:
regular vim doesn't need ~/.vimrc since 7.4, it will happily read ~/.vim/vimrc or so the docs say if (~/.vimrc doesn't exist).

bronzehedwick:
Ack that's right. Can't keep my head straight.

agonzalezro:
The "old" style (`~/vimrc`) is being used for vim: https://github.com/agonzalezro/spf13-vim/blob/patch-1/bootstrap.sh#L124

Sadly, I've clear my history already, but that line was failing. Checking the code is weird because `~/.vim` is a folder linked to `~/.nvim` and the `~/.nvim/.nvimrc` should be allowed, but as I said, the link was failing.

ghost:
and now this PR is obsolete for neovim due to the usage of XDG paths

agonzalezro:
I didn't even remember that this was open. Feel free to close it. Thanks.

PR #829, opened 2015-09-26
Silently restore cursor.

demon386:
I sometimes got "Mark not set" error when opening netrw window for a directory.

vim_error

This ensure that error goes silently when mark is not set.


bronzehedwick:
:+1:

spf13:
Thanks. This one bugged me too.

PR #870, opened 2016-01-03
Remove redundant code

javian:
All this additional code seemed unnecessary.


ghost:
i think the intention here is that program_exists is non-fatal while program_must_exist is fatal.

javian:
Ah I see, I missed that completely. I rewrote it not to remove any existing functionality.

PR #886, opened 2016-02-10
vim-airline themes are now in a dedicated repository, vim-airline-the…

mmikitka:
…mes. See https://github.com/vim-airline/vim-airline/issues/923


howiehu:
+1

PR #910, opened 2016-04-06
Git automatically adds trailing  ^M

ghost:
This can cause some errors for some windows machines. The added command removes it, allowing it to install correctly.


ghost:
this isn't a configuration I'd want to be applied to spf13-vim globally.  I've never used git on windows myself, but maybe this is something you should set in your global git config, but also maybe not. I think some windows users might need to chime in here.

PR #942, opened 2016-06-26
Updated vundle repo

idvoretskyi:
As the actual development of Vundle has been moved to https://github.com/VundleVim/Vundle.vim, updated the reference in the config file.


idvoretskyi:
Invalid

Issue #552, opened 2014-02-06
HTML-AutoCloseTag doesn't work


reem:
Title says it all. I think this might have something to do with autoclose because it makes you "type" the closing > before you actually write the tag type.

Let me know if there's any more info I can provide to help, but I haven't gotten any errors.

bronzehedwick:
+1

I have been using spf13-vim for a few years now, and I have never seen this plugin work. In fact, I only recently found out it was included.

reem:
I've found that it only works if you type the entire tag at once and you don't have any attributes inside the tag.

You have to manually type over the closing > or it doesn't work, and if you move around it doesn't work.

spf13:
It hasn't been updated since 2009. It works, but not well at all. I'm fine with leaving it (it doesn't do any harm) or removing it.

gonzaloserrano:
Totally useless if you ask me :)

chrisnicola:
Currently asking for a github username and password when bundle update is run. I think the repo has disappeared. I'm going to submit a pull request that removes this.

qinyuhang:
There is a third solution, PLZ which one is better? https://github.com/spf13/spf13-vim/pull/1002

PR #999, opened 2016-12-22
Updated HTML-AutoCloseTag

mmjmanders:
amirh's HTML-AutoCloseTag repo doesn't exist anymore

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -244,7 +244,7 @@
 
     " HTML {
         if count(g:spf13_bundle_groups, 'html')
-            Bundle 'amirh/HTML-AutoCloseTag'
+            Bundle 'HTML-AutoCloseTag'
             Bundle 'hail2u/vim-css3-syntax'
             Bundle 'gorodinskiy/vim-coloresque'
             Bundle 'tpope/vim-haml'


yudenzel:
Could we replace it with vim-scripts/HTML-AutoCloseTag ?

chrisnicola:
There have been reports this plugin does not work #552. Can someone verify that it does otherwise we should just remove it as was suggested a couple of years ago.

ProteusCortex:
There is a mirror of the original plugin [here](https://github.com/vim-scripts/HTML-AutoCloseTag), although @amirh's fork had a fix not present in this repo.

ghost:
yes I am having problems with HTML-AutoCloseTag. when I update my plugins, it cannot update the plugin. (using Lnx and vim 8)

ghost:
Does any one have an alternative of the plugin that can be used. I am checking around and wondering if anyone have a suggestion.

kenips:
@chrisnicola I can confirm that HTML-AutoCloseTag still works and this PR fixes the source issue.

conrado:
repositories on https://github.com/vim-scripts don't get curated (PR's looked at, etc) I would recommend using a location like https://github.com/heracek/HTML-AutoCloseTag

PR #1000, opened 2016-12-27
Remove HTML-AutoCloseTag plugin

chrisnicola:
This doesn't plugin doesn't seem supported anymore and the repo is 404'ing. The vim-script repo could be used but people have been reporting this plugin doesn't even work in #552 so I suggest removing it.


conrado (commented):
The plugin does work, perhaps you could consider using a location that is being curated, such as https://github.com/heracek/HTML-AutoCloseTag

PR #1002, opened 2016-12-30
Fix HTML-AutoCloseTag plugin location

pataquets:
No longer hosted at amirh/HTML-AutoCloseTag

### DIFF
--- a/.vimrc.bundles
+++ b/.vimrc.bundles
@@ -244,7 +244,7 @@
 
     " HTML {
         if count(g:spf13_bundle_groups, 'html')
-            Bundle 'amirh/HTML-AutoCloseTag'
+            Bundle 'vim-scripts/HTML-AutoCloseTag'
             Bundle 'hail2u/vim-css3-syntax'
             Bundle 'gorodinskiy/vim-coloresque'
             Bundle 'tpope/vim-haml'


qinyuhang:
Hi, I saw 2 other fix in this GitHub issue, and all you three solution is different from each other... https://github.com/spf13/spf13-vim/issues/552

chrisnicola:
Two of the PRs update the location but don't install it. The one I added removes it as people have reported it doesn't work well anymore. It would be great if a decision could be made.

pataquets:
IMHO, if the plugin is broken (#552), removing it makes sense. Should be confirmed first, I seem to recall it being working when testing spf13-vim some time ago.
Otherwise, if it only has moved to another URL, it should be fixed. Removing something that might be useful for some people, just because of hosting switch seems not OK to me (#999, #1000), unless it is broken.
However, being unmaintained does not matter as long as it works.
Disclaimer: I'm not using it.

pataquets:
Updated above comment and cross-linked.

conrado:
repositories on https://github.com/vim-scripts don't get curated (PR's looked at, etc) I would recommend using a location like https://github.com/heracek/HTML-AutoCloseTag

conrado:
This is a duplicate of #999

acgreek:
some needs to merge this.

PR #1013, opened 2017-01-27
feat(docker): build docker container

ellerbrock:
hey steve,

heads up for all the cool stuff you open sourced on github - respect!
i build a docker container with spf13-vim inside for all those living in the clouds.
you can test it here: <https://hub.docker.com/r/ellerbrock/alpine-spf13-vim/>

docker

have a nice weekend

cheers maik

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,53 @@
+# Developer: Maik Ellerbrock <opensource@frapsoft.com>
+#
+# GitHub:  https://github.com/ellerbrock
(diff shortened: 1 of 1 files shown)


stealthybox:
Thanks for this.
This was a good way to try this setup out.

Seems to run rather slow on my mac, even with a large docker VM.

`docker run -it --rm -v $PWD:/go/src/k8s.io ellerbrock/alpine-spf13-vim`

PR #1049, opened 2017-07-25
.vim is a dir,rm it must use -rf

Pioneer-Linzi:
I use uninstall.sh ,find .vim is dir ,and uninstall use rm and no ``` -rf ``` args , so I add -rf args ,


baiweiguo:
您好！您的邮件已收到，我将会尽快地进行处理，谢谢来信！

PR #1056, opened 2017-08-26
Feature/vim plug

compleatguru:
Hi! I have got [Vim Plug](https://github.com/junegunn/vim-plug) to replace [Vundle](https://github.com/VundleVim/Vundle.vim), mainly for performance due to parallel installing of the plugins.

In addition, I have updated amirh/HTML-AutoCloseTag to vim-scripts/HTML-AutoCloseTag as repo was missing.

Lastly, I have fixed on the uninstall.sh to remove .vim folder and fork files.

Hope this will be useful for the community.

### DIFF
--- a/.vim_plug.unplug
+++ b/.vim_plug.unplug
@@ -1,8 +1,14 @@
 function! s:deregister(repo)
   let repo = substitute(a:repo, '[\/]\+$', '', '')
   let name = fnamemodify(repo, ':t:s?\.git$??')
-  call remove(g:plugs, name)
-  call remove(g:plugs_order, index(g:plugs_order, name))
+
+  try
+    call remove(g:plugs, name)
+    call remove(g:plugs_order, index(g:plugs_order, name))
+  catch /^Vim\%((\a\+)\)\=:E716/ " Key Not present in Dictionary
+    "echom name . ' already removed'
+  endtry
+
 endfunction
 
 command! -nargs=1 -bar UnPlug call s:deregister(<args>)
--- a/.vim_plug.unplug
+++ b/.vim_plug.unplug
@@ -1,0 +1,8 @@
+function! s:deregister(repo)
+  let repo = substitute(a:repo, '[\/]\+$', '', '')
+  let name = fnamemodify(repo, ':t:s?\.git$??')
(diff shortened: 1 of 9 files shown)


VMatrix1900:
You remove the Bundle Unbundle command. How can a user customize the plugin?

compleatguru:
Hi @VMatrix1900 ,
I used [vim-plug](https://github.com/junegunn/vim-plug).
I think just run :PlugClean! to uninstall those unwanted plugins.
You can do more customization if you want to.

What I love about vim-plug is the speed itself, I keep install spf13 on every machine so the faster installation the better for me.

VMatrix1900:
What if I want to remove the plugin shipped with spf13_vim such as nerdcommenter? See 
[https://github.com/spf13/spf13-vim#removing-disabling-an-included-plugin](Readme)

compleatguru:
Maybe this can be of help? https://github.com/junegunn/vim-plug/issues/469

VMatrix1900:
The point is that removing `Unbundle` function is not backward compatible. Can the `Unbundle/Bundle` function implemented using vim-plug? For Unbundle, something like `call remove(g:plus, xxxx)` will do?

compleatguru:
Hi @VMatrix1900 ,
I am sorry but my pull request was to suggest the use of Vim Plug, as you have forked out and perhaps examined the changes, https://github.com/spf13/spf13-vim/pull/1056/files has been modified to use Plug for installation.

As of now the official from vim-plug, https://github.com/junegunn/vim-plug/wiki/tutorial, is to comment on those plugins that you don't need, then run :PlugClean!.

The other alternative solution is to adopt https://github.com/junegunn/vim-plug/issues/469#issuecomment-226965736 where @smancill has contributed an UnPlug function.

Will this be helpful?

compleatguru:
Hi @VMatrix1900 and all,

You can try this out by 
`curl http://bit.ly/spf13_vimplug -L > spf13-vim.sh && sh spf13-vim.sh`
 
I have added UnPlug function.

You may wish to refer [this documentation](https://github.com/compleatguru/spf13-vim/blob/feature/vim-plug/README.vim_plug.md) for example of use.

PR #1121, opened 2021-01-12
Bugfix: for failing Bundle Install

konqui:
if you run command `vim +BundleInstall! +BundleClean +q`
it fails because the Repo amirh/HTML-AutoCloseTag (https://github.com/amirh/HTML-AutoCloseTag) no longer exists.


liming-ai:
could you please tell me how to install them?

